PR #23, opened 2018-11-13
Fixed few callbacks not called

rostopira:
`onIceConnectionState` and `onIceGatheringState` wasn't called when `onSignalingState` is null


rostopira:
Also, this bug could throw null pointer, if only `onSignalingState` is set

Issue #39, opened 2018-12-10
Support volume control for recorded audio


aryzhov:
Since webrtc encapsulates audio capture, it's desirable to control the volume (or gain) of the recorded audio track. I propose to add method `MediaStreamTrack.setVolume(double volume)`, which would apply a gain multiplier to an audio track. The underlying WebRTC API has the corresponding volume control methods, which are platform-specific and exist for both iOS and Android.

rostopira:
LGTM, I will run some tests before merge

rostopira:
Yep, it works. But it's increasing playback volume, not capturer sensitivity
And it's actually gain, not volume, it's even stated in [java doc](https://webrtc.googlesource.com/src/+/master/sdk/android/api/org/webrtc/AudioTrack.java#19)

```java
/** Sets the volume for the underlying MediaSource. Volume is a gain value in the range
--
  | *  0 to 10.
  | */
  | public void setVolume(double volume) {
  | nativeSetVolume(getNativeAudioTrack(), volume);
  | }
```

Maybe will be better to rename method to `setGain`?

cloudwebrtc:
I think named ```setVolume ``` is fine, there is no ambiguity for a single AudioTrack.

rostopira:
@cloudwebrtc okay, I will merge then

PR #40, opened 2018-12-10
#39: Implemented MediaStreamTrack.setVolume()

rostopira:
Thanks @aryzhov !

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,15 @@
+*.iml
+.idea
 .DS_Store
 .packages
 .vscode/.DS_Store
 example/pubspec.lock
 pubspec.lock
 example/ios/Podfile.lock
+build
+GeneratedPluginRegistrant.java
+example/android/.gradle
+example/android/gradle*
+WorkspaceSettings.xcsettings
+example/.flutter-plugins
+example/android/local.properties
--- a/android/src/main/java/com/cloudwebrtc/webrtc/FlutterWebRTCPlugin.java
+++ b/android/src/main/java/com/cloudwebrtc/webrtc/FlutterWebRTCPlugin.java
@@ -269,6 +269,11 @@ public class FlutterWebRTCPlugin implements MethodCallHandler {
             String trackId = call.argument("trackId");
             mediaStreamTrackSwitchCamera(trackId);
             result.success(null);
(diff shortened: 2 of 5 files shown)


aryzhov:
@rostopira Dima, I appreciate the quick response. However, there's an issue with my commit. The code executes but I see no effect in the output volume - neither on iOS, nor on Android. Do you have suggestions on how to proceed?

aryzhov:
Yes, I am using it on local streams, and that's how I thought it's supposed to work. Let me try to change it to remote stream. Also, please take a look at the change I made to FlutterRTCMediaStream.m. It populates localTracks to be later used in setVolume. I have a feeling I might not be doing things correctly there.

rostopira:
@aryzhov Yeah, I've seen this, but that line doesn't do anything useful
If you check [media_stream.dart](https://github.com/cloudwebrtc/flutter-webrtc/blob/master/lib/media_stream.dart) you can see, that `getMediaTracks()` actually populates tracks lists, so you can later call `getAudioTracks()` to get them, so make sure you call it at least once in dart
Maybe I'm wrong about that, because obj-c code in this repo needs some refactoring

aryzhov:
So how do I get a remote stream? `RTCPeerConnection` has `List<MediaStream> _remoteStreams` which never gets populated.

rostopira:
@aryzhov this code should do the trick
```dart
_peerConnection.onAddStream((stream) async {
  await stream.getMediaTracks();
  stream.getAudioTracks().forEach((track) => track.setVolume(10.0));
});
```

aryzhov:
Thanks, that works. Should I try adding code to populate _remoteStreams in order to make it convenient so change the volume during the session, not just to set it initially? Otherwise, the client has to keep its own list of audio tracks.

rostopira:
@aryzhov I'm actually keeping list of streams, and are you sure, that `_remoteStreams` never populated? 🤔

aryzhov:
Yes, keeping a list of streams would be better. `_remoteStreams` (as well as `_localStreams`) don't not get populated, I double-checked. They either need to be removed, or there needs to be logic added to populate them. For now, I'll add one more commit to undo the unnecessary changes in `FlutterRTCMediaStream.m`, as well as to clean up the log messages I added.

PR #58, opened 2019-02-02
Migrated to AndroidX using Refactoring from Andoid Studio

aryzhov:
Since Flutter 1.1.8 was released, it created a bit of a havoc in the Flutter community because of migrating to [AndroidX](https://developer.android.com/jetpack/androidx/). AndroidX replaces the Google Android support libraries, which has been an issue because of version compatibility. Here are some of the open issues related to AndroidX support that I came across as someone who uses a lot of Flutter plugins in a single app:
[Plugins are broken since move to AndroidX](https://github.com/flutter/flutter/issues/27106)
[Plugins do not support AndroidX](https://github.com/flutter/flutter/issues/23995)

Flutter-webrtc happens to be one of the plugins that's causing issues since migrating to AndroidX. While there's an interesting workaround for utilizing non-AndroidX plugin which involves setting `android.enableJetifier=true`, that workaround is not perfect and it doesn't address the core issue which is the inability of different versions of support libraries to co-exist. The right thing to do is to migrate all Flutter plugins to AndroidX. This commit is exactly that. I performed automatic migration by using the Refactoring method described in this document: [Migrating to AndroidX](https://developer.android.com/jetpack/androidx/migrate). 

I tested my changes by running the example project on iOS and Android.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -26,7 +26,7 @@ android {
 
     defaultConfig {
         minSdkVersion 18
-        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
     }
 
     lintOptions {
@@ -41,5 +41,5 @@ android {
 
 dependencies {
     api 'org.webrtc:google-webrtc:1.0.26131'
-    implementation "com.android.support:support-annotations:22.0.0"
+    implementation "androidx.annotation:annotation:1.1.0-alpha01"
 }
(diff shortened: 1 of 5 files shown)


cloudwebrtc:
@aryzhov 
Thanks for you PR, I have added you to collaborators. :)

LinusU:
Would it be possible to cut a release of this? ☺️ 

I can't build my app for Android currently since some other plugins I'm using (the first-party Firebase plugins from Google) have migrated...

rostopira:
@LinusU You can always specify git dependency instead of pub
```
  flutter_webrtc:
    git: https://github.com/cloudwebrtc/flutter-webrtc.git
```
I can't push to pub, only @cloudwebrtc has permission to do that

PR #65, opened 2019-03-21
Fix indentation in peer connection factory

LinusU:
Seems like the entire function after line 9 was indentated by an extra level, this patch fixes that and some other minor whitespace things ☺️


LinusU:
Thanks for the quick pull 👏 🚀

Issue #72, opened 2019-04-08
RTCVideoRenderer.onFirstFrameRendered never called on iOS


rostopira:
**To Reproduce**
```
final ren = RTCVideoRenderer();
    await ren.initialize();
ren.onFirstFrameRendered = () => print("First frame rendered");
```
**Expected behavior**
\*sigh\* -_- 
**Platform information**
iOS
* **Flutter version**: 1.3.8
* **Plugin version**: master branch

PR #73, opened 2019-04-08
Fix onFirstFrameRendered never called on iOS #72

rostopira:


rostopira:
@cloudwebrtc I don't know ¯\_(ツ)_/¯
Otherwise, I wouldn't request a review
Edit this PR as you wish

cloudwebrtc:
I thought about it, it should be needed. If multiple video viewers need to be switched multiple times in the case of multiple video renderers, onFirstFrameRendered should be triggered when the new track is set.

rostopira:
@cloudwebrtc committed, is it fine now?

cloudwebrtc:
@rostopira 
Yes, no problem, I will merge immediately.

PR #76, opened 2019-04-17
Fix self property access in video renderer

LinusU:
Fixes the following error:

```
error: property '_isFirstFrameRendered' not found on object of type 'FlutterRTCVideoRenderer *'; did you mean to access instance variable '_isFirstFrameRendered'?
            self._isFirstFrameRendered = false;
                ~^
```


rostopira:
Yeah, my fault, thanks for fix

PR #86, opened 2019-05-09
Avoid encode/decode overhead with b64 strings for type binary

SimpliciTea:
We are using this in a performance sensitive capacity and wanted to avoid using base64 strings, which get encoded/decoded more than once, when the data type passed by Flutter's `StandardMethodCodec` for typed lists, `FlutterStandardDataType` has a `data` property that can be pulled off and passed to `RTCDataBuffer`'s alloc call directly. This also lets us avoid encoding/decoding our raw binary on the Dart side to b64 in order to use these data channels.


SimpliciTea:
Would it be alright if I just added `.vscode` wholesale to the .gitignore?

SimpliciTea:
Canceling this and moving PR to our master branch

SimpliciTea:
I've just realized that these changes are iOS specific and have not tested for android, so this is not ready to merge until I can find some time to verify nothing is broken, unless anybody else is interested in taking that on.

cloudwebrtc:
Yes, we can wait for the android part of the test to merge without problems, maybe we can write an demo of the binary transfer in the examples, before I tested the string type data transfer in [flutter-webrtc-demo/blob/master/lib/src/call_sample/data_channel_sample.dart](https://github.com/cloudwebrtc/flutter-webrtc-demo/blob/master/lib/src/call_sample/data_channel_sample.dart).

cloudwebrtc:
It seems that we should create a DataBuffer class, defined by a constructor to be string or binary, for example: DataBuffer.fromBinary(Uint8List data), DataBuffer(String sting);

```
class DataBuffer {
  dynamic _data;
  bool _isBinary;

  DataBuffer(String string) {
    this._data = string;
    this._isBinary = false;
  }
  DataBuffer.fromBinary(Uint8List binary) {
    this._data = binary;
    this._isBinary = true;
  }

  bool get isBinary => _isBinary;

  String get string => _data;

  Uint8List get binary => _data;
}
```

This way we can limit the data type to String or Uint8List anyway.

rostopira:
Yeah, `DataBuffer` approach would be definitely better
And my 5 cents about android: this will create a lot of overhead due to passing buffer from native to java and back to native, unless top 1 requested feature is implemented: https://github.com/flutter/flutter/issues/7053
I don't need this feature right now, but if I would, then passing string path and handling it's transfer on java side would be a better option, but it's way too different from original webrtc

cloudwebrtc:
I will merge it first. Later we may need to do the same processing for Android parts by removing ``` if (Platform.isAndroid){``` and using ```java.lang.String```, ```bytes[]``` to maintain the consistency of the plugin communication layer.
Regarding the annotation style, it may be handled together when the final document is online, https://github.com/cloudwebrtc/flutter-webrtc/issues/36, don't worry, it's good now.

PR #88, opened 2019-05-11
Change datachannel message type identifier from a string to an enum, and remove it altogether from the send method and message listener.

FrobtheBuilder:
Map lookups should be faster than string comparisons, so this will likely improve performance.
This will change the API (again) but it will also make it safer, since the compiler will notify you if you mistype the identifier as an enum.

I also did some slight reformatting elsewhere in the file.

### DIFF
--- a/example/lib/src/data_channel_sample.dart
+++ b/example/lib/src/data_channel_sample.dart
@@ -54,15 +54,15 @@ class _DataChannelSampleState extends State<DataChannelSample> {
   /// Send some sample messages and handle incoming messages.
   _onDataChannel(RTCDataChannel dataChannel) {
     dataChannel.onMessage = (type, message) {
-      if (type == "text") {
+      if (type == MessageType.text) {
         print(message.text);
       }
       else {
         // do something with message.binary
       }
     };
-    dataChannel.send("text", RTCDataChannelMessage("Hello!"));
-    dataChannel.send("binary", RTCDataChannelMessage.fromBinary(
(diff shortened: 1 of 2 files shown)


cloudwebrtc:
@FrobtheBuilder @rostopira @Codeusa
I am wondering if we can remove the enumeration type from the interface because the ```RTCDataChannelMessage``` class has internally defined the type to binary or text.
Then the ```send``` method only uses the unique parameter ```message```, and the native sdk provides only text and binary types, so I only need to determine the type when constructing the DataBuffer.
```
Future<void> send(RTCDataChannelMessage message) async {
    ...
}
```

For these we can refer to WebRTC native:
https://chromium.googlesource.com/external/webrtc/+/branch-heads/m74/sdk/objc/api/peerconnection/RTCDataChannel.h#126

https://chromium.googlesource.com/external/webrtc/+/branch-heads/m74/sdk/android/api/org/webrtc/DataChannel.java#160

Another change is that the base64 type I think can be removed, because base64 is essentially a string, just encode before the transfer, decode it after receiving it, now we have the binary type, base64 is unnecessary.

Codeusa:
It does seem redundant to put the message type on the interface itself. So long as the type is set via a map internally on the DataBuffer/RTCDataChannelMessage it should be efficient. String comparisons are indeed slow though so the idea of this change is good. 

The usefulness of the base64 type IMO is that data channels do support arbitrary strings, and base64 is a work around for the binary data on Android for now. So there needs to be a way to tell the two apart until the Android side is brought to spec and either has just binary data or a string.

Issue #87, opened 2019-05-10
Exception on flutter v1.5


cloudwebrtc:
Running an exception on flutter v1.5, it seems that we need to use Main Thread to callback the result of MethodChannel/EventChannel.

```
W/System.err( 5858): java.lang.RuntimeException: Methods marked with @UiThread must be executed on the main thread. Current thread: signaling_threa - 5899
W/System.err( 5858):    at io.flutter.embedding.engine.FlutterJNI.ensureRunningOnMainThread(FlutterJNI.java:605)
W/System.err( 5858):    at io.flutter.embedding.engine.FlutterJNI.dispatchPlatformMessage(FlutterJNI.java:515)
W/System.err( 5858):    at io.flutter.embedding.engine.dart.DartMessenger.send(DartMessenger.java:74)
W/System.err( 5858):    at io.flutter.embedding.engine.dart.DartExecutor.send(DartExecutor.java:149)
W/System.err( 5858):    at io.flutter.view.FlutterNativeView.send(FlutterNativeView.java:142)
W/System.err( 5858):    at io.flutter.plugin.common.EventChannel$IncomingStreamRequestHandler$EventSinkImplementation.success(EventChannel.java:223)
W/System.err( 5858):    at com.cloudwebrtc.webrtc.PeerConnectionObserver.sendEvent(PeerConnectionObserver.java:332)
W/System.err( 5858):    at com.cloudwebrtc.webrtc.PeerConnectionObserver.onRenegotiationNeeded(PeerConnectionObserver.java:425)
```

rostopira:
It's not the only place where it crashes, it's a lot of calls everywhere
If it was Kotlin, I would create extension method like `.successOnUiThread()`, something like that
But in Java wrapping every single single in static method... That would be ugly as hell
Any ideas on how this should be implemented?

rostopira:
Suggestion, wrap every `Result` object in this:
```java
public final class AnyThreadResult implements MethodChannel.Result {
    final private MethodChannel.Result result;
    
    AnyThreadResult(MethodChannel.Result result) {
        this.result = result;
    }
    
    @Override
    public void success(Object o) {
        final Handler handler = new Handler(Looper.getMainLooper());
        handler.post(()->result.success(o));
    }

    @Override
    public void error(String s, String s1, Object o) {
        final Handler handler = new Handler(Looper.getMainLooper());
        handler.post(()->result.error(s, s1, o));
    }

    @Override
[...]

PR #93, opened 2019-05-17
Patches for #87

rostopira:
Fixes issue #87

### DIFF
--- a/android/src/main/java/com/cloudwebrtc/webrtc/utils/AnyThreadResult.java
+++ b/android/src/main/java/com/cloudwebrtc/webrtc/utils/AnyThreadResult.java
@@ -7,6 +7,7 @@ import io.flutter.plugin.common.MethodChannel;
 
 public final class AnyThreadResult implements MethodChannel.Result {
     final private MethodChannel.Result result;
+    final private Handler handler = new Handler(Looper.getMainLooper());
 
     public AnyThreadResult(MethodChannel.Result result) {
         this.result = result;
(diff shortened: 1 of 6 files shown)


rostopira:
Maybe we should check, if already on main thread?
Like that:
```java
    @Override
    public void success(Object o) {
        final Looper main = Looper.getMainLooper();
        if (Looper.myLooper() != main) {
            final Handler handler = new Handler(main);
            handler.post(() -> result.success(o));
        } else
            result.success(o);
    }
```

android/src/main/java/com/cloudwebrtc/webrtc/utils/AnyThreadResult.java
```diff
@@ -0,0 +1,32 @@
+package com.cloudwebrtc.webrtc.utils;
+
+import android.os.Looper;
+import android.os.Handler;
+
+import io.flutter.plugin.common.MethodChannel;
+
+public final class AnyThreadResult implements MethodChannel.Result {
+    final private MethodChannel.Result result;
```
cloudwebrtc: I think we can move the handler to a member variable so that we can reduce duplicate rows, if this doesn't throw any exceptions.
```final Handler handler = new Handler(Looper.getMainLooper());```

rostopira: Yeah, good idea, I will change that

rostopira: Maybe even static final?

cloudwebrtc: If it doesn't throw an exception, I think it's okay. :)

cloudwebrtc: I am not sure how the android thread works, assuming that different threads call the same handler.post at the same time.

postacik:
Hi, will you generate a new version (0.1.7) for this?

cloudwebrtc:
@postacik  pub has been updated to 0.1.7.

postacik:
Thanks 👍

postacik:
FYI

I had exceptions after I switched to v0.1.7.

After executing` flutter clean` the exceptions went away.

PR #101, opened 2019-06-04
Microphone mute

wbarnard:
Adds support for microphone mute feature in Android/iOS.


MorochoRochaDarwin:
Could you provide an example?. How to disable  the microphone when the user click on one button?

wbarnard:
> Could you provide an example?. How to disable the microphone when the user click on one button?

Using the variable naming in the sample code you could call `_localStream?.getAudioTracks()[0].setMicrophoneMute(true)` to disable the microphone.

PR #109, opened 2019-06-26
fix(Peer-connections) split dispose and close methode 

GuillaumeJulien:
ref : https://github.com/cloudwebrtc/flutter-webrtc/issues/106


cloudwebrtc:
LGTM

PR #114, opened 2019-08-12
[Android] Fix 'createIceServer' method Invalid value error

looffee:
I fixed an issue when IceServerMap has an array of url strings and also username and credentials.


cloudwebrtc:
LGTM

PR #129, opened 2019-10-26
Add method for setting torch state

LinusU:
This currently only implements the iOS side of this. There seems like there could be a way to get this to work on Android, but it's a bit hacky since it involves keeping a modified copy of the `CameraSession` class from the upstream WebRTC project here. If this is something you would interested in I'd be happy to look into it, but we would be happy just to have this on iOS ☺️ 

ref: https://stackoverflow.com/a/56656932/148072


cloudwebrtc:
Cool.

PR #145, opened 2019-11-26
Android torch support

LinusU:
I have tested this on a BlackView A10 (which uses the Camera1 API) and @MrAlek have tested this on a Samsung Galaxy A40 (which uses the Camera2 API)


cloudwebrtc:
LGTM

PR #146, opened 2019-11-27
Fix Android frame orientation

LinusU:
Opening this for early feedback. The `rotate*` functions are horribly broken right now because I don't know anything about YUV 😄 

I'll probably get time to complete this next week, but I'm away on other things for the few coming days now. Just wanted to open it in case someone else is also working on this.

links:
- https://developer.android.com/reference/android/graphics/YuvImage
- https://stackoverflow.com/questions/15131663/how-to-change-orientation-of-captured-byte-frames-through-onpreviewframe-callb
- https://stackoverflow.com/questions/14167976/rotate-an-yuv-byte-array-on-android
- https://stackoverflow.com/questions/23107057/rotate-yuv420-nv21-image-in-android
- http://www.wordsaretoys.com/2013/10/25/roll-that-camera-zombie-rotation-and-coversion-from-yv12-to-yuv420planar/
- https://stackoverflow.com/questions/31859708/rotate-yuv420sp-image-by-90-degrees-counter-clockwise
- https://chromium.googlesource.com/libyuv/libyuv/+show/master/docs/rotation.md

### DIFF
--- a/android/src/main/java/com/cloudwebrtc/webrtc/record/FrameCapturer.java
+++ b/android/src/main/java/com/cloudwebrtc/webrtc/record/FrameCapturer.java
@@ -1,6 +1,9 @@
 package com.cloudwebrtc.webrtc.record;
 
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
 import android.graphics.ImageFormat;
+import android.graphics.Matrix;
 import android.graphics.Rect;
 import android.graphics.YuvImage;
 import android.os.Handler;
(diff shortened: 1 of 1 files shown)


LinusU:
This is currently reading up the jpeg, rotating it, and then writing it back. It's of course not the most efficient implantation, but I since it's not in a hot path I don't think that it's a huge issue. If I have some time in my spare time I should be able to resurrect the [raw YUV rotation](https://github.com/cloudwebrtc/flutter-webrtc/compare/master...LinusU:android-frame-orientation-yuv).

ping @cloudwebrtc, ready for review ☺️

cloudwebrtc:
LGTM

PR #152, opened 2019-12-06
Fix render issue

hicodeboy:
Fixed rendering issues caused by remote MediaStream using the same msid/label when using multiple PeerConntions.


cloudwebrtc:
LGTM

PR #188, opened 2020-02-09
Replace Libyuv dependency, fixing #168

### LINKED ISSUE #168: iOS: Incompatibility with FirebaseFirestore because of libyuv headers
See discussion in: https://github.com/firebase/firebase-ios-sdk/issues/4537

Because of flattened headers inside `libyuv.h` coming from https://github.com/waitingsnow/libyuv-iOS/blob/master/headers/libyuv.h the iOS pod of `flutter-webrtc` has become incompatible with `FirebaseFirestore` on iOS.

The `waitingsnow` mirror of libyuv is two years behind upstream and isn't actively maintained but one could create a new fork to fix the issue https://chromium.googlesource.com/libyuv/libyuv/+/refs/heads/master/include/libyuv.h

The question is who will maintain that mirror. If @cloudwebrtc could, that would be amazing.

MrAlek:
Fixes #168 

Created my own fork which uses same binary as https://github.com/waitingsnow/libyuv-iOS/ with updated headers. Tested on example project + our production app.

### DIFF
--- a/ios/flutter_webrtc.podspec
+++ b/ios/flutter_webrtc.podspec
@@ -15,7 +15,7 @@ A new flutter plugin project.
   s.source_files = 'Classes/**/*'
   s.public_header_files = 'Classes/**/*.h'
   s.dependency 'Flutter'
-  s.dependency 'libyuv-iOS'
+  s.dependency 'Libyuv', '1703'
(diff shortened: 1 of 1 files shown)


cloudwebrtc:
LGTM

PR #203, opened 2020-03-11
enabled addTrack API for android and ios

yonatann:
I have enabled the new addTrack API for ios and android as well as added a few runtime checks for robustness


cloudwebrtc:
LGTM

carlfost:
Since addStream() is obsolete, and you add the addTrack(), how do you use it since pc.addTrack(audiotrack,stream) gives syntax error.

yonatann:
The way to use it is to send a list of stream Ids not a stream object: `pc.addTrack(track, [stream.id]);`

vladimircores:
When this API will be available in the release build? Thanx.

animeshjain:
+1 @vladimircores 
Any idea when this will be released?

mapkbalaji:
Please release this API. I'm unable to connect to iOS devices using addStream.

PR #215, opened 2020-04-03
[iOS] Fix typo in render size change check condition

ycherniavskyi:
Currently `didTextureChangeVideoSize` event generates permanently when iOS device in portrait orintation.


cloudwebrtc:
LGTM

PR #223, opened 2020-04-11
Improve RTC video view

ycherniavskyi:
The main intent of this PR is to add to `RTCVideoRenderer` possibility to notify of its changes not only to `RTCVideoView` by direct handler by to all concerned code logic.

For example, this gives possibly to change the size of widget which on its turn holds `RTCVideoView` widget.

Also as the improvement part, `objectFit` and `mirror` were moved from `RTCVideoRenderer` where they only stored to `RTCVideoView` where they actually applied.


ycherniavskyi:
@cloudwebrtc, if you have any considerations or additional proposals, I am ready to adapt my PR.

ycherniavskyi:
Ups, just notice inconsistency with already existed `dispose` method of `RTCVideoRenderer` and method with the same name in `ChangeNotifier`. I am trying to eliminate it shortly.

cloudwebrtc:
@ycherniavskyi I have invited you to collaborator list.

ycherniavskyi:
@cloudwebrtc thank you for the invite! I am going to finish this PR in the next few days and then ask you for your review.

philipgiuliani:
Hi, how are you doing with this PR? :) Looks like a very good improvement.
Currently as workaround I am changing the `key` of the Widget whenever I want to change the `mirror` option for example.

philipgiuliani:
Hi! I tested this branch and it worked great. I only had a problem with the initial rendering of the stream.
The view does not show the image from the camera until I do a hot-reload.

wer-mathurin:
Hi @ycherniavskyi,

there is a conflicting file.  Do you think you can resolve the issue and resubmit the PR.  I think this is a really good improvement and others improvements can be made on top of your proposal.

This will make it easy for @cloudwebrtc to merge it, if it make sense for him.  What is you tough on this one?

lib/rtc_video_view.dart
```diff
@@ -6,19 +6,51 @@ import 'media_stream.dart';
...
+  const RTCVideoValue({
+    this.width = 0.0,
+    this.height = 0.0,
+    this.rotation = 0,
+  });
+
+  double get aspectRatio {
+    if (width == 0.0 || height == 0.0) {
+      return 1.0;
+    } else {
```
serjoca, line 26: The `else` is unnecessary here since there's already a return statement in the `if` above it.

wer-mathurin:
I'll be interested into seeing how you have done the hybrid view for mobile
to web.

Is it something you can share with us?

serjoca:
Sure @wer-mathurin. The only drawback is that it needs to clip the transform since it can overflow a little to the bottom, but it works pretty well and I use that to render a view that is capable of rendering a portrait/landscape/square based on the renderer's own **aspect ratio**.



```dart
import 'dart:math';

import 'package:flutter/foundation.dart';
import 'package:flutter/material.dart';
import 'package:flutter_webrtc/webrtc.dart';

class VideoRenderer extends StatefulWidget {
  final double width;
  final double height;
  final RTCVideoRenderer renderer;

  VideoRenderer({
    @required this.width,
    @required this.height,
    @required this.renderer,
  });

  @override
  _VideoRendererState createState() => _VideoRendererState();
}

class _VideoRendererState extends State<VideoRenderer> {
  @override
  Widget build(BuildContext context) {
    bool isRendererAvailable =
        widget.renderer != null && widget.renderer.textureId != null;
    final size = MediaQuery.of(context).size;

    return isRendererAvailable
        ? kIsWeb
            ? Container(
                width: widget.width,
                height: widget.height,
                color: Colors.black,
                child: RTCVideoView(widget.renderer),
              )
            : Container(
                color: Colors.black,
                child: ClipRRect(
                  child: Transform.scale(
                    scale: size.width >= size.height
                        ? size.aspectRatio / widget.renderer.aspectRatio
                        : widget.renderer.aspectRatio / size.aspectRatio,
                    child: Center(
                      child: AspectRatio(
                        aspectRatio: widget.renderer.aspectRatio,
                        child: Transform(
                          transform: Matrix4.identity()..rotateY(-pi),
                          alignment: FractionalOffset.center,
                          child: Texture(
                            filterQuality: FilterQuality.medium,
                            textureId: widget.renderer.textureId,
                          ),
                        ),
                      ),
                    ),
                  ),
                ),
              )
        : Container();
  }
}
```

cloudwebrtc:
@serjoca You can try my last commit.

ycherniavskyi:
I am sorry that I didn't finish my PR, and thankful that you have done it! 👍

nalbion:
Mirror does not work for in Chrome on Windows

```dart
                RTCVideoView(
                  _localRenderer,
                  mirror: true,
                )
```

nalbion:
If I request a small QVGA stream, I'm given a tall skinny video. It's easily scalable in the aspect ratio that it's provided in, but is it possible to crop/clip/mask the video so that it's not so tall?

PR #228, opened 2020-04-17
[macOS] Fix hot restart videoCapturer crash

artsector:
If RTCVideoView with opened _localRenderer placed at root Widget, after hot restarting application was crashed with error:

*** Assertion failure in -[RTCCameraVideoCapturer dealloc], ../../sdk/objc/components/capturer/RTCCameraVideoCapturer.m:113
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Session was still running in RTCCameraVideoCapturer dealloc. Forgot to call stopCapture?'

### DIFF
--- a/macos/Classes/FlutterRTCMediaStream.m
+++ b/macos/Classes/FlutterRTCMediaStream.m
@@ -290,6 +290,9 @@ typedef void (^NavigatorUserMediaSuccessCallback)(RTCMediaStream *mediaStream);
     
     if (videoDevice) {
         RTCVideoSource *videoSource = [self.peerConnectionFactory videoSource];
+        if(self.videoCapturer){
+            [self.videoCapturer stopCapture];
+        }
         self.videoCapturer = [[RTCCameraVideoCapturer alloc] initWithDelegate:videoSource];
         AVCaptureDeviceFormat *selectedFormat = [self selectFormatForDevice:videoDevice];
         NSInteger selectedFps = [self selectFpsForFormat:selectedFormat];


cloudwebrtc:
LGTM

macos/Classes/FlutterRTCMediaStream.m
```diff
@@ -290,6 +290,9 @@ - (void)getUserVideo:(NSDictionary *)constraints
     
     if (videoDevice) {
         RTCVideoSource *videoSource = [self.peerConnectionFactory videoSource];
+        if(self.videoCapturer){
```
ycherniavskyi, line 293: Hm, I see inconsistent with previous code formating here ☹️ - whitespace before and after the condition is absent. But, unfortunately, PR already merged. @cloudwebrtc could I add formatting fixing commit directly to master?

cloudwebrtc: @ycherniavskyi No problem, you can fix it directly.

ycherniavskyi:
@artsector do I understand it correct that the same fix must be introduced for iOS version also (I mean [here](https://github.com/cloudwebrtc/flutter-webrtc/blob/5c3852a8e060dc187b52ea25e1edc4dc6edfe755/ios/Classes/FlutterRTCMediaStream.m#L293-L294))?

artsector:
@ycherniavskyi ios restarts without any crashes, but yes it is a same issue

PR #231, opened 2020-04-19
Example as demo

truongsinh:
I did not know why the example was not used as the full demo, so I moved all the latest changes there in to the `example` dir, with further
- factoring
- proper way to handle URL
- a known free TURN server credential
- `flutter format .`


cloudwebrtc:
@truongsinh I do n’t think we need to put the complete example in the plugin, because most people will directly rely on the flutter-webrtc plugin to develop their own products. Similar to the p2p sample, the data channel needs to connect to a specific server, which will make people mistake this It is a plug-in bundled with a specific server.
The plug-in itself only requires the most basic API testing and loopback support, and the other is to make the plug-in code as simple as possible.

truongsinh:
OK thanks, then if I have some ideas to improve the demo, I will do in the other repo instead.

ycherniavskyi:
@cloudwebrtc maybe we could create the wiki page where collect such type of additional demos when users commit them to some public repo?

@truongsinh I am a bit curious, but what additional features of [flutter-webrtc](https://github.com/cloudwebrtc/flutter-webrtc) usage does your demo add to compare to [flutter-webrtc-demo](https://github.com/cloudwebrtc/flutter-webrtc-demo)? Maybe you could create PR for that project?

cloudwebrtc:
> @cloudwebrtc maybe we could create the wiki page where collect such type of additional demos when users commit them to some public repo?

This is a good idea.

PR #238, opened 2020-04-24
error when videoRenderer dispose

yungzhu:
When the video is not turned on
textureId is null


ycherniavskyi:
But maybe it is better to add this check to `dispose` method of `RTCVideoRenderer` class in Dart so that it will apply to all implementations. What do you think?

yungzhu:
> But maybe it is better to add this check to `dispose` method of `RTCVideoRenderer` class in Dart so that it will apply to all implementations. What do you think?

I re-pushed a modification.
https://github.com/cloudwebrtc/flutter-webrtc/pull/256

wer-mathurin:
@yungzhu: After all changes done to the library, do you think this PR is still needed?

PR #240, opened 2020-04-24
Make optional 'constraints' argument in createPeerConnection()

shepeliev:
As we have default constraints in _createPeerConnection()_ we could make that arg optional in order to avoid passing empty map as following:

```dart
final pc = await createPeerConnection(configuration);
```

instead of: 
```dart
final pc = await createPeerConnection(configuration, {});
```


cloudwebrtc:
LGTM!

PR #251, opened 2020-05-12
make createOffer constraints optional

Zangetsu101:


ycherniavskyi:
API modification is backward compatible and corresponds to [`createOffer`](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/createOffer) method of `RTCPeerConnection` API, so LGTM (as @cloudwebrtc writes 😊).

PR #256, opened 2020-05-18
Fix dispose error

yungzhu:
Fix the problem of calling dispose immediately when initialize is not completed


wer-mathurin:
@yungzhu :  Have you tried the latest version of the library to see if the problem occur, and if this is the case, can you resolve the conflict and resubmit the PR.

Issue #259, opened 2020-05-20
Multiple RTCVideoRenderer objects on the same Flutter Web HTML page


leonidl:
When having multiple RTCViewRenderer objects on the same HTML page, srcObject can be attached to a wrong HTML.VideoElement. This happens since findHtmlView() method cannot distinguish between different HTML video elements and will return an incorrect element instead of null.

Another issue related to findHtmlView() can happen when another Flutter Web Widget uses ui.platformViewRegistry.registerViewFactory() method. In this case there will be an exception since findHtmlView() element will find an HTML element and it will not be of type HTML.ViewElement.

I am going to attach a PR once the issue is created.

leonidl:
PR #260 is submitted.

PR #260, opened 2020-05-20
Fixing multiple video renderers on the same HTML page for Flutter Web

leonidl:
This commit fixes the case where multiple RTCVideoRenderer objects present on the same Flutter Web HTML page and where another object was using ui.platformViewRegistry.registerViewFactory() method.

This PR is supposed to fix issue #259 .


cloudwebrtc:
LGTM

PR #263, opened 2020-05-22
Add peerConnectionId to data channel EventChannel

### LINKED ISSUE #262: Wrong EventChannel with multiple RTCDataChannels and PeerConnections
With multiple peer connections and DataChannels, the wrong `onMessage` and `onDataChannelState` gets called. This is because the `EventChannel` name only depends on the `dataChannelId`. But it should depend on the `peerConnectionId` as well.

optimygmbh:
Fix #262


cloudwebrtc:
LGTM

PR #316, opened 2020-08-24
Added label as read only property

maanas:
The https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel specification provides read-only label for the data channel. Added a getter to get the label


cloudwebrtc:
LGTM

PR #330, opened 2020-09-05
Fix - NPE if onIceGatheringState is not defined

wer-mathurin:
FIX - missing null check onIceGatheringState (web)
REFACTOR - add final keyword
REFACTOR - remove new keyword
REFACTOR - replace if statement by null check operator for readability
FIX - Adding stream twice to the _remoteStream list


cloudwebrtc:
LGTM

PR #337, opened 2020-09-11
cleaning and fix potential concurrency issue

wer-mathurin:
continue cleaning
remove duplicate call to notifyListeners() -> already call in the setter
fix potential concurrency issue


cloudwebrtc:
LGTM

PR #338, opened 2020-09-11
continue cleaning and code robustness

wer-mathurin:
Web: Reuse code to prevent copy and paste code
Web: Remove confusing print, move to the API comment
Mobile: Add null check to prevent passing null to native implementations


cloudwebrtc:
LGTM

cloudwebrtc:
@wer-mathurin Sorry, I have refactored web/rtc_video_view.dart to solve the problem of video disappearing when resizing the window and simplify the code.

PR #344, opened 2020-09-14
Let's add a first test

wer-mathurin:
An attempt to add a first test to validate that not setting any public delegate will not throw a NPE


cloudwebrtc:
LGTM

cloudwebrtc (approved):

PR #350, opened 2020-09-17
Issues/349

wer-mathurin:
The bug was coming from that all StreamSubscription that was call after the renderer was dispose.

Note:
I tried adding a test, but since it basically always succeed.(see related bug)  Maybe I'm not calling the test the right way !


cloudwebrtc:
LGTM

cloudwebrtc (approved):

PR #356, opened 2020-09-26
Updated rtc_dtmf_sender

maanas:
The api call is name insertDTMF
https://developer.mozilla.org/en-US/docs/Web/API/RTCDTMFSender/insertDTMF
renamed to function call to match with the WebRTC Api.

### DIFF
--- a/example/lib/src/loopback_sample.dart
+++ b/example/lib/src/loopback_sample.dart
@@ -179,7 +179,7 @@ class _MyAppState extends State<LoopBackSample> {
   void _sendDtmf() async {
     var dtmfSender =
         _peerConnection.createDtmfSender(_localStream.getAudioTracks()[0]);
-    await dtmfSender.sendDtmf('123#');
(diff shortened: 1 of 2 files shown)


cloudwebrtc:
@maanas Thanks, renaming is good, the example also needs to be renamed, otherwise `flutter analyze` will report an error.
https://github.com/flutter-webrtc/flutter-webrtc/blob/master/example/lib/src/loopback_sample.dart#L182

maanas:
@cloudwebrtc Thanks, I should have renamed the example. I have left _sendDtmf the function call as it is. As that depends on the user. Hope that is OK.
I will update the https://github.com/maanasnetwork/flutter_janus once these changes are pushed.

  ```
  // pluginHandle.dtmfSender.insertDTMF(tones, duration, gap);
    pluginHandle.dtmfSender
        .sendDtmf(tones, duration: duration, interToneGap: gap);
```

cloudwebrtc:
LGTM

PR #357, opened 2020-09-26
RTCDTMFSender class doesnt need to MediaStreamTrack 

xyzbilal:
Because MediaStreamTrack is not used by RTCDTMFSender class argument removed from createDtmfSender() function in rtc_peerconnection.dart


cloudwebrtc:
https://github.com/flutter-webrtc/flutter-webrtc/blob/master/lib/src/web/rtc_peerconnection.dart#L238

We must keep the web and native code consistent, so I tried the same interface model as js, and later I will directly use MediaStreamTack and RtpSender (on unified-plan) to create DTMFSender.

xyzbilal:
I get the point. I am closing it.

PR #364, opened 2020-10-02
WIP - api standardization across implementation

wer-mathurin:
This work can be seen as a first step to standardize API accros implementation.  It prevent missing method implementation, as an example, in the master branch hasTorch() have been implemented for native but not Web.  If you use this method and try to run it on when you will encounter compilation failure.

There is also others scenarios that also happen.  But merging this PR, this resolve all those cases.

You will find in the code two way of exposing the API.  The first one is just exporting that interfaces, and delegate the instanciation of the real class implementation to methods exposed by a factory.

The second one is exporting a class that implement the interface and use a delegate, see RTCVideoRenderer.dart for the example.

I personally like the second option because if we eventually move the the plugin federation concept if we will only need to copy/paste classes in the right package!  Because everything if organize and code like plugin federation, but in one plugin :-)


cloudwebrtc:
@wer-mathurin There are two file conflicts, I can try to merge after the conflict is resolved.

wer-mathurin:
Conflict resolved

lib/src/interface/rtc_dtmf_sender.dart
```diff
@@ -1,9 +1,4 @@
-import 'dart:html' as html;
-
-class RTCDTMFSender {
-  RTCDTMFSender(this._jsDtmfSender);
-  final html.RtcDtmfSender _jsDtmfSender;
-
+abstract class IRTCDTMFSender {
```
cloudwebrtc, line 1: rename `IRTCDTMFSender` => `RTCDTMFSender`.

lib/src/interface/media_recorder.dart
```diff
@@ -0,0 +1,22 @@
+import 'enums.dart';
+import 'media_stream.dart';
+import 'media_stream_track.dart';
+
+abstract class IMediaRecorder {
```
cloudwebrtc, line 5: rename `IMediaRecorder` => `MediaRecorder`.

lib/src/web/factory_impl.dart
```diff
@@ -0,0 +1,59 @@
+import 'dart:async';
+import 'dart:convert';
+import 'dart:html' as html;
+
+import 'package:flutter_webrtc/src/interface/rtc_dtmf_sender.dart';
```
cloudwebrtc, line 5: Here should be imported using a relative path.
```dart
import '../interface/rtc_dtmf_sender.dart';
import 'rtc_dtmf_sender_impl.dart';
```

lib/src/web/rtc_video_view_impl.dart
```diff
@@ -0,0 +1,58 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_webrtc/src/web/rtc_video_renderer_impl.dart';
```
cloudwebrtc, line 2: Here should be imported using a relative path.

```dart
import 'rtc_video_renderer_impl.dart';
```

cloudwebrtc (changes requested):

cloudwebrtc (approved):

cloudwebrtc:
LGTM

wer-mathurin:
@rostopira Can you look at the code and give comments.  This is a big change so I'm looking for your input.

Issue #365, opened 2020-10-11
Example error on master branch


samuelg123:
**Describe the bug**
I found error in the example.
<img width="775" alt="Screenshot 2020-10-11 155202" src="https://user-images.githubusercontent.com/12625711/95674956-f0c43300-0bdd-11eb-8e5a-5d8b0c5f849a.png">

**To Reproduce**
Just run example at master branch and choose GetUserMedia API Test.

**flutter doctor -v**: 
```
[✓] Flutter (Channel beta, 1.22.1, on Microsoft Windows [Version 10.0.19041.508], locale en-ID)
    • Flutter version 1.22.1 at C:\src\flutter\beta
    • Framework revision f30b7f4db9 (3 days ago), 2020-10-08 10:06:30 -0700
    • Engine revision 75bef9f6c8
    • Dart version 2.10.1

[✓] Chrome - develop for the web
    • Chrome at C:\Program Files (x86)\Google\Chrome\Application\chrome.exe
    • Flutter plugin version 50.0.1
    • Dart plugin version 193.7547
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b01)
```

I just made a PR #366 

Thank you

PR #366, opened 2020-10-11
Fix error expected a value of type 'RTCVideoRendererWeb'

samuelg123:
This will fix issue #365


test/unit/web/rtc_videw_view_test.dart
```diff
@@ -1,12 +1,12 @@
 @TestOn('browser')
 import 'package:flutter_test/flutter_test.dart';
-import 'package:flutter_webrtc/flutter_webrtc.dart';
+import 'package:flutter_webrtc/flutter_webrtc.dart' hide RTCVideoRenderer;
```
cloudwebrtc, line 3: I don’t quite understand why `hide RTCVideoRenderer` here

cloudwebrtc:
@samuelg123 fixed https://github.com/flutter-webrtc/flutter-webrtc/commit/975a9f00fe37e7e32d6b3ecfc23b9f47e340126c

samuelg123:
Great. I will close this.

test/unit/web/rtc_videw_view_test.dart
wer-mathurin: This an error.  But this test is not working anyway.  I will write a working test when the below issue will be resolve.
The annotation @TestOn is not working.  see https://github.com/flutter/flutter/issues/66045

Issue #400, opened 2020-11-03
PeerConnection removeTrack throws `Sender is null` error, MediaStream removeTrack throws `Track is null`.


Ayman-Kortobaa:
**Describe the bug**
In unified-plan when calling `mediaStream.removeTrack(sender.track)` it throws track is null error because the track id is null.
And also in `peerConnection.removeTrack(sender)` it throws sender is null issue.

**To Reproduce**
```dart

videoSender =
              await peerConnection.addTrack(track, [signaling.mediaStream]);
await signaling.mediaStream.removeTrack(videoSender.track);
await peerConnection.removeTrack(videoSender);
```

**Expected behavior**
Removing the tracks successfully.

* **Flutter doctor**:
```
[√] Flutter (Channel stable, 1.22.2, on Microsoft Windows [Version 10.0.19041.572], locale en-US)
    • Flutter version 1.22.2 at C:\src\flutter
    • Framework revision 84f3d28555 (3 weeks ago), 2020-10-15 16:26:19 -0700
    • Engine revision b8752bbfff
    • Dart version 2.10.2

[√] Android toolchain - develop for Android devices (Android SDK version 29.0.3)
    • Android SDK at C:\Users\pc\AppData\Local\Android\Sdk
    • Platform android-30, build-tools 29.0.3
    • ANDROID_HOME = C:\Users\pc\AppData\Local\Android\Sdk
    • Java binary at: C:\Program Files\Android\Android Studio\jre\bin\java
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b01)
    • All Android licenses accepted.

[√] Android Studio (version 4.0)
    • Android Studio at C:\Program Files\Android\Android Studio
    • Flutter plugin version 48.1.2
    • Dart plugin version 193.7547
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b01)

[√] VS Code (version 1.50.1)
    • VS Code at C:\Users\pc\AppData\Local\Programs\Microsoft VS Code
    • Flutter extension version 3.16.0

[√] Connected device (1 available)
    • Redmi Note 9 Pro (mobile) • 7d3b60de • android-arm64 • Android 10 (API 29)

• No issues found!
```
* **Plugin version**: `master` branch
* **OS**: Android
* **OS version**: 10

PR #401, opened 2020-11-03
[Android] Fixed `Track is null` 'Sender is null' issues when removing track.

Ayman-Kortobaa:
This pull request fixes the following issues:
In unified-plan when calling `mediaStream.removeTrack(sender.track)` it throws track is null error because the track id is null as in the Java code it is called `trackId` while in Dart code it anticipates `id`.
And also in `peerConnection.removeTrack(sender)` it throws sender is null issue because in the java code the track id is compared with the `==` operator which compares reference instead of equality, it is replaced with the `.equals()` method to fix this issue.
I have created the issue here https://github.com/flutter-webrtc/flutter-webrtc/issues/400


cloudwebrtc:
LGTM

Issue #403, opened 2020-11-05
[iOS] Not found video track for RTCMediaStream: {streamId} error when using Unified-Plan `onTrack` and videos don't render.


Ayman-Kortobaa:
**Describe the bug**
Using `onTrack` of `PeerConnection` results in a `Not found video track for RTCMediaStream: {streamId}` error in the console and videos not being rendered in UI.

**To Reproduce**

```dart
peerConnection.onTrack = (event) async {
        final stream = event.streams.first;
        final track = event.track;
        if (track.kind == 'video') {
            final remoteRenderer = RTCVideoRenderer();
            await remoteRenderer.initialize();
            remoteRenderer.srcObject = stream;
            _remoteRenderers.putIfAbsent(stream.id, () => remoteRenderer);
          }
    };
```

**Expected behavior**
The error should not be thrown as there are video tracks and they should renderer normally.

Issue #402, opened 2020-11-05
[iOS]  The Method '[ ]' was called on null "result"  when calling `PeerConnection`'s `removeTrack`


Ayman-Kortobaa:
**Describe the bug**
Calling `PeerConnection`'s `removeTrack` method in Unified-Plan throws `The Method '[ ]' was called on null "result" ` in Dart.

**To Reproduce**
```dart
final videoSender = await peerConnection
          .addTrack(videoTracks.first, [signaling.mediaStream]);
await peerConnection.removeTrack(videoSender);
```

**Expected behavior**
The track is probably removed successfully as this error is in the Dart code, but it should complete successfully and return a boolean whether it was removed or not.

PR #404, opened 2020-11-05
[iOS] fix removeTrack methodChannel response, onTrack's `stream` and `track` not being registered in native.

Ayman-Kortobaa:
This pull request is for two issues:
1. The `removeTrack`method in `rtc_peerconnection_impl.dart` anticipates a boolean in `Map` with `result` as a key
```dart
	bool result = response['result'];
```

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; but the `removeTrack` in Darwin's sends `nil` which throws `The Method '[ ]' was called on null "result"` in Dart.

2. The second issue was not registering `stream` and `track` in `onTrack` for `PeerConnection` - `didAddReceiver` in `Obj-c` - which results into a `Not found video track for RTCMediaStream: {streamId}` error.

______________________
Related issues: 
1. https://github.com/flutter-webrtc/flutter-webrtc/issues/402
2. https://github.com/flutter-webrtc/flutter-webrtc/issues/403

### DIFF
--- a/common/darwin/Classes/FlutterRTCPeerConnection.m
+++ b/common/darwin/Classes/FlutterRTCPeerConnection.m
@@ -545,7 +545,10 @@ didStartReceivingOnTransceiver:(RTCRtpTransceiver *)transceiver {
                 }
             }
         }
-        
+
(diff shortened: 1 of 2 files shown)


cloudwebrtc:
LGTM

Issue #405, opened 2020-11-06
Unsupported value error when using `setParameters` on `RTCRtpSender` due to  `mapToRtpParameters` not being implemented.


Ayman-Kortobaa:
**Describe the bug**
When using unified-plan and calling `setParameters` on `RTCRtpSender` an `unsupported value` error gets thrown because it is being sent `null` and `nil` on Android and iOS respectively due to `mapToRtpParameters` not being implemented.

**Expected behavior**
The function should set and update RtpParameters correctly.

PR #407, opened 2020-11-08
[Darwin/Android] `RtpSender` `setParameters` functionality.

Ayman-Kortobaa:
This pull request fixes the following issues:
In unified-plan when calling `RTCRtpSender.setParameters(parameters)` it throws `Unsupported value` error because the value being sent is `null` as in native code `mapToRtpParameters` returns `null`.
Creating `RtpParameters` from a map is not easy as there is no constructor for some of its properties, and setting the parameter is mainly targeted towards encodings.
So `mapToRtpParameters` is replaced with `updateRtpParameters` to update the values of the sender's parameters (encodings only as the rest is read-only) with the values sent from Dart through the method channel.

Related issue: https://github.com/flutter-webrtc/flutter-webrtc/issues/405


cloudwebrtc:
LGTM

PR #413, opened 2020-11-15
Fix adding track to stream 

igala:
The method tried to find the tracks in the localTRacks map instead of parsing peerconnection remote track


cloudwebrtc:
LGTM

cloudwebrtc:
LGTM

PR #414, opened 2020-11-16
fix java string compare

yuroller:
While debugging transceiver api in android, I found two small bugs in string compares.
There are still some problems in transceiver api, but this is a start.


cloudwebrtc:
LGTM

Issue #390, opened 2020-10-25
Using the track received from ontrack in unified plan


igala:
How should a track be played when received.

I tried adding it to a stream created locally but it fails.

Any suggestions?

Ayman-Kortobaa:
Can you share your code?
Audio tracks are handled on the native side, the `stream` containing Video `track`s has to be added to a `RTCVideoRenderer`'s `srcObject` then viewed in `RTCVideoView`.
The example in the package can help you get started.

igala:
Below is a code snippet from janus.js from the onTrack callback under the unified plan branch:

`config.pc.ontrack = function(event) {
				Janus.log("Handling Remote Track");
				Janus.debug(event);
				if(!event.streams)
					return;
				config.remoteStream = event.streams[0];
				if(!event.track)
					return;
				// Notify about the new track event
				var mid = event.transceiver ? event.transceiver.mid : event.track.id;
				try {
					pluginHandle.onremotetrack(event.track, mid, true);
`

What goes back with the callback is track and mid.

on the application side it handles the track by cloning the track and adding it to a new stream created locally, then played with video view. Below is the snippet:

`stream = new MediaStream();
					stream.addTrack(track.clone());
					remoteFeed.remoteTracks[mid] = stream;`


How do you accomplish this with flutter webrtc?

cloudwebrtc:
@igala https://github.com/flutter-webrtc/flutter-webrtc/blob/master/example/lib/src/loopback_sample.dart#L91

igala:
Thank you,

I followed to code example and at first it looks like it worked , but entering a room with multiple video tracks an using the stream received in the ontrack event, shows only one stream.

The stream received in the ontrack event holds the same stream for all mids.

It looks like a bug.

itsthetaste:
I think I'm hitting that same issue.

I can have managed to get the mobile to web working, but the web to mobile isn't.

Inside my web js, inside my connection, I'm initially creating an inactive transceiver for both the audio and video.
Then when i want to send either audio or video, I do a getUserMedia, and replace the tracks in the appropriate transceivers and set their direction to sendonly, then negotiate.

In my flutter app, I'm seeing the onTrack fire twice, once for the audio (I ignore this as that handled internally), and then again with the video track.

For me though, the streams list is empty, so I cannot go and place it into the srcObject of my renderer.

This is what I can see in the debugger.
[...]

PR #420, opened 2020-11-27
Unified plan gettracks

itsthetaste:
when the remote end is using stage 3 of the evolution of tracks (see https://blog.mozilla.org/webrtc/the-evolution-of-webrtc/) and using transceivers.  

`var t = pc.addTransceiver("video", { direction : 'inactive'});`

and then at a later time adding a track to the sender

```
t.direction = "sendonly";
t.sender.replaceTrack(track);
```

Streams are not used so I needed to add this code so as to be able to display the track in the video renderer.

I've had to make a local stream, add the track to it, which meant going and finding it in the transceiver.  Then the video renderer can take the new local stream and play the track.


itsthetaste:
#390

cloudwebrtc:
LGTM

PR #422, opened 2020-11-30
[WIP] - Selecting an audio output

wer-mathurin:


cloudwebrtc:
LGTM

cloudwebrtc (approved):

wer-mathurin:
Do we want to merge this pull request without the Native implementation?
I tried an implementation for set muted and get muted.

But there is a  throw UnimplementedError in the class RTCVideoRendererWeb for set audiooutput

cloudwebrtc:
Oops, I should add the native implementation, please don't merge it for now.

wer-mathurin:
After looking at the Android(java) implementation, getSource is not returning audiooutputs

Also I saw that is returning a weird key for audioinput.  As decribe by the MediaDeviceInfo API, we should have
**kind, label, groupId and deviceId** 

The only thing we have right now is a constant for audio output !
ConstraintsMap audio = new ConstraintsMap();
    audio.putString("label", "Audio");
    audio.putString("deviceId", "audio-1");
    audio.putString("facing", "");
    audio.putString("kind", "audioinput");

I saw there is an Interface AudioManager and there is anonymous implementation in FlutterWebRTCPlugin.  Maybe you can add two methods, getAudioInput and getAudioOutputs and the delegate RTCAudioManager can return the values?

cloudwebrtc:
@wer-mathurin What do you think of changing `audioOutput` to `setSinkId` and `deviceId` to `sinkId`?

wer-mathurin:
The sinkId refer to the method (this is still experimental).
But the Helper audioOutputs return an MediaDeviceInfo, like the audioInputs.

The way to call it on Web (only for Web) 
await audio.setSinkId(audioDevices[0].deviceId);

The function still need an deviceId. 

I personally think if we want to provide an API over Native and Web, we must keep the smallest denominator, and sinkId is only use for Web and this is more human friendly.

On the other side, since we only support Web for now, maybe it can make sense too :-0

I'd propose to merge it as is, and get back to it from the community comments.

cloudwebrtc:
LGTM

Issue #201, opened 2020-02-28
Bluetooth headset switching on iOS


arcticfox1919:
I wonder if there are any plans to switch Bluetooth headsets on iOS? I read the source code and it's implemented in Android, but there is no such code on iOS.

cloudwebrtc:
Welcome patches.

romilmeghani:
+1 for this functionality. When can it be expected?

Ayman-Kortobaa:
Is there any news about this issue?

PR #435, opened 2020-12-21
Bluetooth switching enabled when switching `enableSpeakerphone`  value (if they are connected). #201

Ayman-Kortobaa:
This is a quick fix to enable Bluetooth switching on iOS and fix the issue where setting `enableSpeakerphone` to false would default to the phone (call) speaker even if a Bluetooth headset is connected.
This would fix the #201 issue, we aren't able to pick devices though.


cloudwebrtc:
LGTM

Issue #461, opened 2021-01-29
setMicrophoneMute - Not working properly


efraespada:
Some problems has appeared in the last update (`v0.5.8`).

Android: `setMicrophoneMute` doesn't produce any change. The microphone is always muted.
iOS: `setMicrophoneMute` has been inverted. I mean: 

This enables the microphone:
```dart
track.setMicrophoneMute(true);
```

This mutes the microphone:
```dart
track.setMicrophoneMute(false);
```

**To Reproduce**
Do any call and try to switch the mute state of the microphone (Android or iOS).

**Expected behavior**
`track.setMicrophoneMute` method working properly like in prior versions (`v0.5.7`).

**Platform information**
Any Android or iOS device (tested in multiple physical devices).

* **Flutter version**:
```
[✓] Flutter (Channel stable, 1.22.5, on macOS 11.1 20C69 darwin-x64, locale en-ES)
    • Flutter version 1.22.5 at /Users/efraespada/Documents/flutter
    • Framework revision 7891006299 (7 weeks ago), 2020-12-10 11:54:40 -0800
    • Engine revision ae90085a84
    • Dart version 2.10.4
```
* **Plugin version**: `0.5.8`

* **OS**: Android, iOS
* **OS version**: Android 10 and iOS 14.3

PR #466, opened 2021-02-09
Fix setMicrophoneMute

samuelg123:
Hi,
This will fix issue #461 (inverted microphone mute state). 

Thanks.

Tested on Android 9 & 10.


cloudwebrtc:
LGTM

PR #502, opened 2021-03-23
fix signalingState & iceConnectionState event name

FourLeafTec:
switch signalingState & iceConnectionState event name sink back to flutter on Windows platform


cloudwebrtc:
LGTM

PR #548, opened 2021-04-26
Add factory configuration for Android

danteCarvalho:


cloudwebrtc:
The PR looks good, but I want to know whether this setting is also effective for iOS and macOS and dart API should be effective cross-platform.

danteCarvalho:
Searching for "disableNetworkMonitor" showed this file /macos/WebRTC.framework/Header/RTCPeerConnectionFactoryOptions.h so it should be possible for mac. Ios has the same file in this dependency: s.dependency 'GoogleWebRTC', '1.1.31999'.  But i found nothing for windows.

cloudwebrtc:
Yes, for the current PR, we only need to add this setting to darwin, and the Windows version needs to wait for me to do the package of libwebrtc.

danteCarvalho:
Oh, it seems that since i added a method to the factory.dart we need to add that method to the web build as well.

Issue #551, opened 2021-05-01
Screen Capture Android onResume: Cannot start already started MediaProjection


XuanTung95:
**Describe the bug**
```
E/AndroidRuntime( 8090): FATAL EXCEPTION: main
E/AndroidRuntime( 8090): Process: com.github.pion.ion.ion_flutter_app, PID: 8090
E/AndroidRuntime( 8090): java.lang.RuntimeException: Unable to resume activity {com.github.pion.ion.ion_flutter_app/com.github.pion.ion.ion_flutter_app.MainActivity}: java.lang.IllegalStateException: Cannot start already started MediaProjection
E/AndroidRuntime( 8090): 	at android.app.ActivityThread.performResumeActivity(ActivityThread.java:3805)
E/AndroidRuntime( 8090): 	at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:3845)
E/AndroidRuntime( 8090): 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1773)
E/AndroidRuntime( 8090): 	at android.os.Handler.dispatchMessage(Handler.java:106)
E/AndroidRuntime( 8090): 	at android.os.Looper.loop(Looper.java:164)
E/AndroidRuntime( 8090): 	at android.app.ActivityThread.main(ActivityThread.java:7000)
E/AndroidRuntime( 8090): 	at java.lang.reflect.Method.invoke(Native Method)
E/AndroidRuntime( 8090): 	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:441)
E/AndroidRuntime( 8090): 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1408)
E/AndroidRuntime( 8090): Caused by: java.lang.IllegalStateException: Cannot start already started MediaProjection
E/AndroidRuntime( 8090): 	at android.os.Parcel.readException(Parcel.java:2037)
E/AndroidRuntime( 8090): 	at android.os.Parcel.readException(Parcel.java:1975)
E/AndroidRuntime( 8090): 	at android.media.projection.IMediaProjection$Stub$Proxy.start(IMediaProjection.java:140)
E/AndroidRuntime( 8090): 	at android.media.projection.MediaProjection.<init>(MediaProjection.java:69)
E/AndroidRuntime( 8090): 	at android.media.projection.MediaProjectionManager.getMediaProjection(MediaProjectionManager.java:165)
[...]

PR #552, opened 2021-05-01
Do not call reStartCamera() for screen capture

XuanTung95:
1.  Fix bug** #551
- Add variable VideoCapturerInfo->isScreenCapture.
- Do not call reStartCamera in case of screen capture

2. Add options "cryptoOptions" when create peerconnection so user can enable Gcm crypto which is much faster.


cloudwebrtc:
LGTM

cloudwebrtc:
@XuanTung95 Can support `cryptoOptions` for iOS/macOS? I think other platforms also need consistency.

cloudwebrtc:
Has been added in this commit.
https://github.com/flutter-webrtc/flutter-webrtc/commit/d3eaf1698fa7c302fd41215be20905efeb20991e#diff-2496b27d4acaf6c7dc7482b1b221ccc92d20f720d1ca50d9aaffa404f0306564R1218

XuanTung95:
> @XuanTung95 Can support `cryptoOptions` for iOS/macOS? I think other platforms also need consistency.

Idk, I only develop on Android.

Issue #565, opened 2021-05-18
[iOS] Example app crashes on iOS due to incorrect syntax in `RTCConfiguration` method


Ayman-Kortobaa:
**Describe the bug**
The application crashes with this error `-[__NSDictionaryM objectAtIndexedSubscript:]: unrecognized selector sent to instance 0x2826ac7e0`.

**To Reproduce**
Start the example application on iOS and start a call (loopback example for instance).

**Expected behavior**
The PeerConnection should initiate correctly with correct configuration and the call should start.

NOTE: This happens on master branch not a version uploaded to Pub.dev

PR #566, opened 2021-05-18
[Darwin] Fix incorrect syntax in private api in `RTCConfiguration` #565

Ayman-Kortobaa:
Fix incorrect syntax in private api in `RTCConfiguration` in `FlutterWebRTCPlugin.m` causing app crashes in Darwin devices.

This should fix issue #565


cloudwebrtc:
LGTM

PR #581, opened 2021-06-05
[Android] falling back to the first available camera fix #580

### LINKED ISSUE #580: getUserMedia() fails on some Android devices (TVs with cameras)
**Describe the bug**
Getting the error `Unable to getUserMedia: getUserMedia(): Failed to create new track` on certain Android devices that include cameras or can have external cameras like TVs.

**To Reproduce**
Start a video meeting on Android TV like Huawei IdeaHub, you will get the error above.

**Expected behavior**
The application should fall back to the first available camera.

Ayman-Kortobaa:
Added code to fall back to the first available camera in case the `createVideoCapturer` fails to find a suitable camera based on constraints (doesn't have a front/back camera) like an Android TV that will most likely have an external camera or any other type.

### DIFF
--- a/android/src/main/java/com/cloudwebrtc/webrtc/GetUserMediaImpl.java
+++ b/android/src/main/java/com/cloudwebrtc/webrtc/GetUserMediaImpl.java
@@ -279,7 +279,13 @@ class GetUserMediaImpl {
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


cloudwebrtc:
LGTM

PR #585, opened 2021-06-07
Add label getter to DataChannel Interface

### LINKED ISSUE #584: Add Label getter to DataChannel Interface
Currently, it is not possible to get a data channel's label via the RTCDataChannel Interface. It would be useful to have this method during discovery of data channels in the onDataChannel Callback.

rlshumat:
Adds a label getter to the DataChannel Interface and implements the label getter for the web implementation.

Closes #584


cloudwebrtc:
LGTM

PR #588, opened 2021-06-09
[Web] fix: exception raises at RTCPeerConnection and RTCRtpSender

jjpark-edito:
Web implementation of `RTCPeerConnection .removeTrack` and `RTCRtpSender.setParameters` raises following exception. 

> Expected a value of type 'FutureOr<bool>', but got one of type 'Null' 

Actually, at the javascript implementation, type of the return value of these two functions is `void`, but implementation of `RTCPeerConnection .removeTrack` and `RTCRtpSender.setParameters` are assuming that the return type is `bool`.

To prevent this exception, I changed these two functions to ignore the return value came from javascript version of the function and just return a `true`.


cloudwebrtc:
LGTM

PR #631, opened 2021-07-14
[web] Switch transferFromImageBitmap to be invoked using js.callMethod

### LINKED ISSUE #621: Example not working: captureFrame function in GetUserMedia API test not working in release mode
**Describe the bug**
The GetUserMedia API test capture frame does not work in release mode although it works in debug mode.
The error that I am getting is:
image
The issue is caused by the line: 
final frame = await videoTrack.captureFrame();

**To Reproduce**
1: Run the example application in release mode in Chrome : flutter run --release
2: Navigation from: GetUserMedia API Test -> Click call button -> Click capture image button

**Expected behavior**
A dialog with captured image is expected to show up.

**Platform information**

* **Flutter version**: Channel stable, 2.2.2
* **Plugin version**: 0.6.5
* **OS**: Mac OS Chrome
* **OS version**: 10.15.7
* **Chrome version**: 91.0.4472.114 (Official Build) (x86_64)

kenoung:
The current implementation of captureFrame works in debug mode but fails in
profile and release mode with `TypeError: ... is not a function`. This commit
uses js.callMethod to invoke the method instead of using the native api, which
seems to resolve this issue.

Fixes #621

### DIFF
--- a/lib/src/web/media_stream_track_impl.dart
+++ b/lib/src/web/media_stream_track_impl.dart
@@ -67,7 +67,7 @@ class MediaStreamTrackWeb extends MediaStreamTrack {
     canvas.height = bitmap.height;
     final renderer =
         canvas.getContext('bitmaprenderer') as html.ImageBitmapRenderingContext;
-    renderer.transferFromImageBitmap(bitmap);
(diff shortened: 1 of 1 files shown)


cloudwebrtc:
LGTM

PR #634, opened 2021-07-14
Fix sending binary data over DataChannel in web implementation.

### LINKED ISSUE #633: Sending binary RTCDataChannelMessage from flutter-web to Android fails to deliver payload
I have successfully established connections; if I run two Android instances it works fine. If I run one flutter-web and one Android and send an `RTCDataChannelMessage.fromBinary(...)` both ways, the one from Android is delivered successfully, but the one from the flutter-web instance arrives as an RTCDataChannelMessage with isBinary=true and binary=[0,0,0,0,0,0,...]

I have tried printing the message data immediately before sending and immediately upon receiving, and this failure is fully reflected - the 'about to send' binary data is not all zeroes, and the received binary data, only in the failure case, is. It also doesn't match the length - sending a 14-byte packet triggers the receiving code, and the resulting msg.binary of all zeroes is too long for the log message to contain.

As further confirmation that something is going wrong in the binary sending or receiving code somewhere in the RTC library's handling, if I change the sending end from
```
    channel.send(RTCDataChannelMessage.fromBinary(msg));
```
to
```
    channel.send(RTCDataChannelMessage((const Base64Encoder()).convert(msg)));
```

And the receiving end from
```
final binary = msg.binary;
```
to
```
final binary = (const Base64Decoder()).convert(msg.text);
```

Then the message is delivered successfully.

I am going to use this workaround for now, since it works, but I would much prefer to send the binary data!

ravenblackx:
Fixes #633.

### DIFF
--- a/lib/src/web/rtc_data_channel_impl.dart
+++ b/lib/src/web/rtc_data_channel_impl.dart
@@ -58,9 +58,7 @@ class RTCDataChannelWeb extends RTCDataChannel {
     if (!message.isBinary) {
       _jsDc.send(message.text);
     } else {
-      // This may just work
-      _jsDc.sendByteBuffer(message.binary.buffer);
-      // If not, convert to ArrayBuffer/Blob
+      _jsDc.sendTypedData(message.binary);
     }
     return Future.value();
   }


cloudwebrtc:
LGTM

PR #660, opened 2021-08-27
web: properly cleanup srcObject to avoid accidental dispose

### LINKED ISSUE #659: web: mediaTrack stopped when renderer disposed
**Describe the bug**
when the renderer gets destroyed, it'll inadvertently destroy the attached MediaStream (and the tracks that they contain).

**To Reproduce**
```dart
renderer.srcObject = null;
renderer.dispose();
```

this will cause dispose to be called on MediaStreamWeb, stopping all tracks in the stream. This is undesirable as a different UI element may need to render the track.

**Expected behavior**
after srcObject is set to null, dispose no longer impact it.

* **Plugin version**: 0.6.5
* **OS**: web

davidzhao:
fixes #659


cloudwebrtc:
LGTM

PR #667, opened 2021-09-01
web: avoid removing all audio elements when stopping a single video renderer

davidzhao:
When a single video renderer is removed, it removes the entire audio manager div, causing all audio tracks to stop playback.


cloudwebrtc:
LGTM

PR #670, opened 2021-09-02
Web related fixes

hiroshihorie:
- Added `MediaStreamTrack.getSettings` for Web
- `RTCPeerConnection.addTransceiver` for Web uses `track` instead of `kind` when both are specified.


cloudwebrtc:
LGTM

PR #675, opened 2021-09-03
Web: `addTransceiver` bug fix

hiroshihorie:
fix bug: 
```dart
init is RTCRtpTransceiverInitWeb
```
 **is always null** so init will never be set.

### DIFF
--- a/lib/src/web/rtc_rtp_transceiver_impl.dart
+++ b/lib/src/web/rtc_rtp_transceiver_impl.dart
@@ -49,7 +49,6 @@ class RTCRtpTransceiverInitWeb extends RTCRtpTransceiverInit {
 }
 
 extension RTCRtpTransceiverInitWebExt on RTCRtpTransceiverInit {
-  //
   dynamic toJsObject() => jsutil.jsify({
         'direction': typeRtpTransceiverDirectionToString[direction],
         if (streams != null)
--- a/lib/src/web/rtc_rtp_transceiver_impl.dart
+++ b/lib/src/web/rtc_rtp_transceiver_impl.dart
@@ -39,26 +39,21 @@ class RTCRtpTransceiverInitWeb extends RTCRtpTransceiverInit {
         listToRtpEncodings(map['sendEncodings']));
   }
 
-  Map<String, dynamic> toMap() {
-    return {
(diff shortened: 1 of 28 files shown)


lib/src/web/rtc_rtp_transceiver_impl.dart
```diff
@@ -45,15 +46,18 @@ class RTCRtpTransceiverInitWeb extends RTCRtpTransceiverInit {
         if (sendEncodings != null)
           'sendEncodings': sendEncodings!.map((e) => e.toMap()).toList(),
       };
+}
 
-  Map<String, dynamic> toJSMap() => {
+extension RTCRtpTransceiverInitWebExt on RTCRtpTransceiverInit {
+  //
```
cloudwebrtc, line 52: I think the comment mark can be removed to keep the code tidy.

cloudwebrtc (changes requested):

lib/src/web/rtc_rtp_transceiver_impl.dart
hiroshihorie: Sure!

cloudwebrtc: I did a test on android devices, and I can open simulcast with a soft encoder on the android simulator, but I cannot open simulcast with a Qualcomm Hardware encoder on Samsung S20.

cloudwebrtc: It can work normally on ios/mac devices.

hiroshihorie: @cloudwebrtc Thanks for testing! Don't know why I can't get it working, do you have a piece of sample code?

cloudwebrtc: https://github.com/pion/ion-sdk-flutter/blob/master/example/lib/views/echotest_view.dart

You can use this code with the latest ion-sfu to use grpc-web signaling, but it will only work on the android emulator.

cloudwebrtc:
LGTM

cloudwebrtc (approved):

Issue #676, opened 2021-09-09
Web no longer works on Flutter 2.5.0


DaveBubenik:
**Describe the bug**
I'm seeing an exception while trying to display a remote stream. The exception originates from the findHtmlView() function in rtc_video_renderer_impl.dart on the following line:
```
final child = (fltPv.first as html.Element).shadowRoot!.childNodes;
```
The above line returns null from shadowRoot, so an exception is thrown. 

Flutter has changed how shadowRoot works in version 2.5.0:
https://flutter.dev/docs/release/breaking-changes/platform-views-using-html-slots-web

**To Reproduce**
Show a remote stream on Web Browser.

**Expected behavior**
No exception to be thrown.

**Platform information**

* **Flutter version**: 2.5.0
* **Plugin version**: 0.6.7
* **OS**: Chrome
* **OS version**: 92.0.4515.159

hiroshihorie:
Yes I'm experiencing this also, trying to figure it out...

LucasMdosAnjos:
Yes I'm experiencing the same situation

hiroshihorie:
Should be fixed with https://github.com/flutter-webrtc/flutter-webrtc/pull/681

carlogrisetti:
Confirm it's fixed by #681

LucasMdosAnjos:
This fix is already in 0.6.7 or will come only in a new version?

carlogrisetti:
It is not in 0.6.7. You need to pull it from master if you need it right now.

PR #681, opened 2021-09-11
Fix: Flutter 2.5.0 `RTCVideoRendererWeb` bug

hiroshihorie:
Should fix this issue https://github.com/flutter-webrtc/flutter-webrtc/issues/676

- Fix: Flutter 2.5.0 `RTCVideoRendererWeb` bug
- Organize imports (Use relative path instead of absolute path of own package)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,31 @@
+name: Build
+
+on:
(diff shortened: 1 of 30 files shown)


cloudwebrtc:
LGTM

PR #693, opened 2021-09-21
add support for lightning microphone

cr0manty:
the problem was with webrtc forcing the AVAudioSessionCategoryPlayAndRecord which overrides the microphone to use the built-in

added force replacement of the category with AVAudioSessionCategoryPlayAndRecord, so that the external microphone can be detected


cloudwebrtc:
LGTM

cloudwebrtc:
It makes sense. If we can solve the lightning microphone and also the issue in https://github.com/webrtc-sdk/webrtc/pull/7#issuecomment-925960331, I will revert this PR.

hiroshihorie:
@cr0manty I recommend you manage audio session at your app level using https://pub.dev/packages/audio_session for example. 😁 
Modifications to audio session at flutter_webrtc level will cause many unintended issues since it's impossible to cover all user's use cases. 😁

Issue #687, opened 2021-09-17
Audio input issue via USB-C/lightning 


cr0manty:
On most devices, the microphone does not work through the usb-c / lightning port, after connecting, or if the microphone was before the start of broadcasting, then audio does not go. And through 3.5 jack the broadcast is going well

external input:
android - not work at all
iphone 12 with apple EarPods iOS 14.7.1 - works fine (audio input from headphone mic) 
<iPhone 12 - not work at all

cloudwebrtc:
This is an interesting question, this should be an issue of audio manager or sound card management. Maybe we should develop an enhanced API to manage various audio devices, wired, wireless, Micro USB, type-c, 3.5mm, Bluetooth, It sounds like there are many things to do. xD

cr0manty:
Yeah, I already start dig into this and it’s really not simple task, some base moves, like handle device notification and set audio on input via `setPreferredInput` Seems not work

cr0manty:
@cloudwebrtc this pull request fix this issue on iOS https://github.com/flutter-webrtc/flutter-webrtc/pull/693

PR #705, opened 2021-09-29
Improve audio handling on iOS

davidzhao:
iOS: enabled audio playback mode without recording
iOS: Fixes audio handling with lightning headphones #687

### DIFF
--- a/common/darwin/Classes/FlutterRTCPeerConnection.m
+++ b/common/darwin/Classes/FlutterRTCPeerConnection.m
@@ -2,8 +2,10 @@
 #import "FlutterWebRTCPlugin.h"
 #import "FlutterRTCPeerConnection.h"
 #import "FlutterRTCDataChannel.h"
+#import "AudioUtils.h"
 
 #import <WebRTC/WebRTC.h>
+
 
 @implementation RTCPeerConnection (Flutter)
 
@@ -170,11 +172,11 @@
 -(void) peerConnectionClose:(RTCPeerConnection *)peerConnection
 {
     [peerConnection close];
-    
+
     // Clean up peerConnection's streams and tracks
     [peerConnection.remoteStreams removeAllObjects];
     [peerConnection.remoteTracks removeAllObjects];
-    
+
     // Clean up peerConnection's dataChannels.
     NSMutableDictionary<NSNumber *, RTCDataChannel *> *dataChannels
     = peerConnection.dataChannels;
@@ -198,9 +200,9 @@
         [peerConnection statsForTrack:track
                      statsOutputLevel:RTCStatsOutputLevelStandard
                     completionHandler:^(NSArray<RTCLegacyStatsReport *> *reports) {
-                        
+
                         NSMutableArray *stats = [NSMutableArray array];
-                        
+
                         for (RTCLegacyStatsReport *report in reports) {
                             [stats addObject:@{@"id": report.reportId,
                                                @"type": report.type,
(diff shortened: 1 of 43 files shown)


common/darwin/Classes/FlutterRTCPeerConnection.m
```diff
@@ -388,19 +390,25 @@ - (void)peerConnection:(RTCPeerConnection *)peerConnection didAddStream:(RTCMedi
...
         [videoTracks addObject:@{@"id": track.trackId, @"kind": track.kind, @"label": track.trackId, @"enabled": @(track.isEnabled), @"remote": @(YES), @"readyState": @"live"}];
     }
-    
+
     NSString *streamId = stream.streamId;
     peerConnection.remoteStreams[streamId] = stream;
-    
+
+    if (hasAudio) {
+        [AudioUtils ensureAudioSessionWithRecording:NO];
```
cloudwebrtc, line 409: When `didAddStream` was triggered, we received a `remote track`, but before that, we had created a `local track` to capture audio. Is there any wrong result here?

davidzhao: that case works fine, `ensureAudioSessionWithRecording` will always upgrade the audio session, and will never downgrade.

* when we received a track: sets to Playback
* then create local track: sets to PlayAndRecord
* then received another track: does nothing

cloudwebrtc: I think there is an easier way:

```
@interface FlutterWebRTCPlugin : NSObject<FlutterPlugin, RTCPeerConnectionDelegate>
...

-(void) ensureAudioSessionWithRecording {
  BOOL hasRecordingTrack = NO;
  for (RTCMediaStreamTrack *track in self.localTracks) {
         if([track isKindOfClass:[RTCAudioTrack class]]) {
                hasRecordingTrack = YES;
         }
   }

 BOOL hasPlaybackTrack = NO;
 for(RTCPeerConnection *pc in self.peerConnections) {
          for (RTCMediaStreamTrack *track in pc.remoteTracks) {
             if([track isKindOfClass:[RTCAudioTrack class]]) {
                    hasPlaybackTrack = YES;
             }
      }
 }
 if(hasRecordingTrack ... hasPlaybackTrack) {
     //do something here.
  }
}
```

davidzhao: sure, that works too!

cloudwebrtc: Then the `ensureAudioSessionWithRecording` method is called after the number of any remote/local track changes.

cloudwebrtc: @davidzhao What do you think about it?

cloudwebrtc (approved):

common/darwin/Classes/FlutterRTCPeerConnection.m
davidzhao: sure, I can update it

davidzhao:
@kNoAPP I just pushed a fix to a new PR #750. Please give it a try locally and let me know if that fixes things for you.

kNoAPP:
@davidzhao thanks. I'll pull the PR and test it shortly!

davidzhao:
@kNoAPP thanks for confirming! I'm glad this fixed it for good.

PR #712, opened 2021-10-04
Fix the bug that turning on or off the speaker does not work

zi6xuan:
[audioSession setPreferredSampleRate:44100.0 error:nil];
After testing, sampleRate must be set to take effect


hiroshihorie:
@davidzhao is this safe to merge ? I'm not sure yet.

hiroshihorie:
I think we need to use `sampleRate` of `RTCAudioSession` instead of directly manipulating AVAudioSession :
https://github.com/webrtc-sdk/webrtc/blob/c85ea8b0e5def03e15f3dfa5399c66bc2f0ecf41/sdk/objc/components/audio/RTCAudioSessionConfiguration.h#L31

I can make a PR for this, what do you guys think @cloudwebrtc @davidzhao ?

cloudwebrtc:
@hiroshihorie I agree, I think you can do this

davidzhao:
I think part of the reason this is broken is my fault. With AudioUtils, it's always preferring speaker:

https://github.com/flutter-webrtc/flutter-webrtc/blob/master/common/darwin/Classes/AudioUtils.m#L20

We'd have to take _speakerOn into consideration.. I suggest moving all audio session related code into the helper class to consolidate.

davidzhao:
tho I'm confused about the current fix about setting sample rate. I think this might be simply by side effect? sample rate shouldn't have anything to do with speaker output device.

hiroshihorie:
@davidzhao You are right, I think RTCAudioSession might be overwriting some settings when `configure` is called internally. So If we just change the code to use `RTCAudioSession` instead of `AVAudioSession` issues that @zi6xuan is experiencing might resolve.

zi6xuan:
After a day of debugging, I found that the AudioSession setting of version 0.6.8 caused the speaker setting button of callkit to fail. I downgraded to 0.6.5 and it seems that there is no problem, but I am facing a new problem. When I click the speaker button, the sound has actually changed from The speaker came out, but the button UI flickered and immediately became non-speaker, I don’t know what went wrong

pauloamarcon:
In version 0.8.0 the same problem returned.

PR #690, opened 2021-09-18
feature: fix Malformed Framework on macOS

efraespada:

Issue #556, opened 2021-05-08
[macos] Error uploading IPA to the AppStore: Malformed Framework.


efraespada:
This error is reported by Apple when uploading the IPA file.

> ITMS-90291: Malformed Framework - The framework bundle WebRTC (Land-a.app/Contents/Frameworks/WebRTC.framework) must contain a symbolic link 'WebRTC' -> 'Versions/Current/WebRTC'. Refer to the Anatomy of Framework Bundles for more information.

> ITMS-90291: Malformed Framework - The framework bundle WebRTC (Land-a.app/Contents/Frameworks/WebRTC.framework) must contain a symbolic link 'Resources' -> 'Versions/Current/Resources'. Refer to the Anatomy of Framework Bundles for more information.

> ITMS-90292: Malformed Framework - The framework bundle WebRTC (Land-a.app/Contents/Frameworks/WebRTC.framework) 'Versions' directory must contain a symbolic link 'Current' resolving to a specific version directory. Resolved link target: '${linkTarget}'. Refer to the Anatomy of Framework Bundles for more information.

**To Reproduce**

Upload any API file with this plugin to the AppStore.

**Platform information**

 [✓] Flutter (Channel stable, 2.0.6, on macOS 11.3.1 20E241 darwin-x64, locale en-GB)
    • Flutter version 2.0.6 at /Users/efraespada/flutter
    • Framework revision 1d9032c7e1 (9 days ago), 2021-04-29 17:37:58 -0700
    • Engine revision 05e680e202
    • Dart version 2.12.3

* **Plugin version**:  ^0.6.3
* **OS**:  macOS
* **OS version**: 11.3.1

cloudwebrtc:
There is a known conflict here. If we use the WebRTC.framework version with soft links, when submitted to pub.dev, all soft links will become actual file uploads, which leads to an increase in the size of the package submitted to pub.dev Several times, there is no problem with local test development, but uploading the app store will encounter the problem you described, you can create a soft link in `~/.pub-cache/hosted/pub.dartlang.org/flutter_webrtc-0.6.5/macos/WebRTC.framework`  the format required by the app store. And try again.

efraespada:
Okay, I'll test it 😉

efraespada:
@cloudwebrtc 

I could not make it work. Could you provide an example of how to do what you say?

Which is different from other plugins that don't have this link problem?

efraespada:
[...]

PR #717, opened 2021-10-06
[WIP] Use pre-compiled WebRTC for macOS.

cloudwebrtc:


cloudwebrtc:
I found a symlink problem in the macOS version of WebRTC.framework. The mac version of WebRTC.framework/WebRTC must use a symbolic link to Versions/Current/WebRTC

So I need to publish a new version to Pod

hiroshihorie:
Oh, really! I will check it out now.

hiroshihorie:
Are you sure the structure is incorrect? because I Archiving succeeds for both ARM & Intel. Also the sample app runs with Rosetta (intel simulated) mode...  (I'm using M1 mac)

スクリーンショット 2021-10-17 3 07 18

cloudwebrtc:
I'm not sure if the swift project has the same problem. This happened in flutter run -d macos, because during code-signing, if the WebRTC.framework/WebRTC file does not use a symbolic link to the subdirectory, it will cause the flutter app of macos Compilation failed.
image

cloudwebrtc:
This PR tries to fix the problem of the old version of the symbolic link
https://github.com/flutter-webrtc/flutter-webrtc/pull/690/files

cloudwebrtc:
Therefore, after your simulcast header file patch is merged, we still need to recompile, and copy the merged WebRTC.framework/WebRTC of the macOS version (intel/m1) to Versions/Current/WebRTC, and create .cxframework again

hiroshihorie:
Ah, you're right something is wrong, looking into it...

cloudwebrtc:
This problem only occurs in the framework signature, but before that, another developer @efraespada  submitted an issue and could not upload the app to the Mac store https://github.com/flutter-webrtc/flutter-webrtc/issues/556. The macOS framework needs to use a symbolic link by default. You can view the WebRTC.framework compiled by webrtc. Before merging into multiple arches, the default is also a symbolic link.

https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/FrameworkAnatomy.html

PR #731, opened 2021-10-19
Enable Android simulcast

hiroshihorie:
Credits to Shiguredo, Inc. Thank you very much @voluntas for the great work. Flutter users will appreciate and benefit from this.
https://github.com/shiguredo/sora-android-sdk/blob/develop/sora-android-sdk/src/main/kotlin/jp/shiguredo/sora/sdk/codec/SimulcastVideoEncoderFactoryWrapper.kt

This PR will turn on Kotlin used by `SimulcastVideoEncoderFactoryWrapper.kt`. I'm not sure if there are any side-effects @cloudwebrtc  pls let me know. @davidzhao @davidliu

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,24 @@
+###################################################################################
+
+The following modifications follow Apache License 2.0 from shiguredo.
+
+SimulcastVideoEncoderFactoryWrapper.kt
+
+Apache License 2.0
+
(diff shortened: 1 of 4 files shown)


hiroshihorie:
Is the size increase huge? I was even thinking we can slowly start to migrate to Kotlin. Which I think long term will have more contributors. Even Swift over Objc for long term I think will have more contributors interested. What do you think? @cloudwebrtc  @davidzhao No need to rush at all tho.

PR #741, opened 2021-10-24
feat: Add support for elinux.

cloudwebrtc:
~Currently, it can be compiled and run, But cross-thread calls will cause a crash, so need debugging.~

Update:
The crash just happens in debug mode, everything works fine in release/profile mode

In debug mode, it seems that the problem occurs with asserts caused by cross-thread calls
````
[FATAL:flutter/fml/memory/weak_ptr.h(111)] Check failed: (checker_.checker).IsCreationThreadCurrent().
````

image


seanchann:
any news?  I think if it is for the arm architecture, does it need to be implemented based on `Gstreamer/webrtcbin` or another portable engineering for embedded os? (current libwebrtc file  is x86 binary )

cloudwebrtc:
Embedded Linux can use v4l2, pulse,  alsa, and the google webrtc library is supported, just need to cross-compile this library
https://github.com/webrtc-sdk/libwebrtc

cloudwebrtc:
@seanchann I've added arm/arm64 binaries in subsequent commits

cloudwebrtc:
replace with  https://github.com/flutter-webrtc/flutter-webrtc/pull/1338

PR #760, opened 2021-11-09
Web: Don't stop tracks when disposing `MediaStream`

hiroshihorie:
I believe the developer should be responsible of stopping tracks if required. Since tracks maybe required for later use.
This matches the behavior with iOS/Android, not calling `stop` on tracks when disposing `MediaStream`.

Additionally, `RTCVideoRendererWeb` shouldn't call `dispose` on `srcObject`. Any time `RTCVideoRendererWeb` gets disposed `srcObject` will get disposed.

I found that `MediaStreamTrack.dispose` is deprecated anyway :
https://github.com/flutter-webrtc/flutter-webrtc/blob/42940fb6f54e1e1ab170e4d66a987383f573d114/lib/src/interface/media_stream_track.dart#L115-L116

I agree `stop` should be called by developer whenever required.

@cloudwebrtc  @davidzhao

### DIFF
--- a/lib/src/web/media_stream_impl.dart
+++ b/lib/src/web/media_stream_impl.dart
@@ -53,14 +53,6 @@ class MediaStreamWeb extends MediaStream {
   }
 
   @override
-  Future<void> dispose() async {
-    getTracks().forEach((element) {
-      element.stop();
-    });
-    return super.dispose();
-  }
-
-  @override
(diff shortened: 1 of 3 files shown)


davidzhao:
this is a good fix. well done!

cloudwebrtc:
LGTM

PR #764, opened 2021-11-15
Properly start foreground service in example

davidliu:
Example app currently only initializes FlutterBackground, and doesn't actually start it.


cloudwebrtc:
LGTM

PR #789, opened 2021-12-01
Fix parsing stun configuration

ashellunts:
Code was expecting a stun value to be under key 'url'.
I have added support for format of stun where url is just a string.

Diff is not good, because most of the code is actually not changed (just inserted inside if block).


hiroshihorie:
@davidzhao @cloudwebrtc Hi guys ! @ashellunts found this issue that was causing a crash on Windows.
I am not fully aware of WebRTC's specs but with this PR both map and string should be ok.

https://developer.mozilla.org/en-US/docs/Web/API/RTCIceServer/urls
https://developer.mozilla.org/en-US/docs/Web/API/RTCIceServer/url

cloudwebrtc:
LGTM

cloudwebrtc (approved):

ashellunts:
@cloudwebrtc can you release new version, please?

PR #790, opened 2021-12-02
New video constraints syntax

davidliu:
Updating getUserMedia to use latest video constraint syntax (https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackConstraints).

Current getUserMedia constraints have been deprecated for years (and documentation is very sparse), so we should start looking to support the latest syntax.

### DIFF
--- a/android/src/main/java/com/cloudwebrtc/webrtc/GetUserMediaImpl.java
+++ b/android/src/main/java/com/cloudwebrtc/webrtc/GetUserMediaImpl.java
@@ -428,16 +428,6 @@ class GetUserMediaImpl {
 
     void getDisplayMedia(
             final ConstraintsMap constraints, final Result result, final MediaStream mediaStream) {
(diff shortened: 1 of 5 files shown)


davidliu:
The PR currently only contains Android changes at the moment, but will update with changes for the other platforms; mainly requesting for comments rather than approval until then. 

The PR only handles the `ideal` property, and won't support min/max right now (as that requires quite a bit more work).

davidliu:
This should now cover all the currently supported platforms (web already handles the latest syntax).

davidzhao:
@cloudwebrtc @hiroshihorie please take a look at this PR. It resolves inconsistencies in media constraint specification. the PR is backwards compatible

hiroshihorie:
Nice update!

cloudwebrtc:
LGTM

cloudwebrtc (approved):

PR #792, opened 2021-12-03
[windows] Fix mute

ashellunts:
There was no response to trackDispose method call.
That causes failure in client code and therefore not working mute.


cloudwebrtc:
LGTM

PR #795, opened 2021-12-05
Fix macOS compile

hiroshihorie:
Recent commits breaks macOS compiling since UIImage is only available on iOS.


cloudwebrtc:
LGTM

PR #796, opened 2021-12-05
macOS screen capturing (video related changes)

hiroshihorie:
* Common interface for video capturers (Camera/Screen) FlutterRTCVideoCapturer
* MacOSScreenCapturer 
* Can have multiple video capturers (previously it's just a single property), so camera & screen share could be used the same time.

I don't know whats `FlutterRTCFrameCapturer` for so I didn't touch that.
FlutterRPScreenRecorder.h/m should be renamed to VideoCapturers.h/m

TODO: make torch compatible


cloudwebrtc:
@hiroshihorie I just escaped from an analyzer conflict. `directives_ordering` conflicted with `import_sorter`, which caused ci to fail after the 0.8.0 tag. now it is fixed, so you need to fetch upstream to `livekit/flutter-webrtc`

cloudwebrtc:
@hiroshihorie @davidzhao  We need to fully consider the screen sharing function of the desktop platform, for example, screen capture is permission/full screen, and the choice of the application window.

First, we need to implement a selection window similar to the following, on the upper layer of the UI of the flutter desktop app. It will pop up when we call getDisplayMedia.

In addition, I need to export and encapsulate the interface in 
https://github.com/webrtc-sdk/webrtc/tree/main/modules/desktop_capture into a suitable API to enumerate screens, application windows, and generate thumbnails.

These native layer interface packages will also be applicable to macOS Windows Linux native SDK.

image

cloudwebrtc:
> I don't know whats FlutterRTCFrameCapturer for so I didn't touch that.

This is used to capture a JPEG image from VideoTrack.

matteoannibali:
Hi, it's possible to test a beta version with this update?

PR #798, opened 2021-12-06
Fix issue with getting remote track

ashellunts:
In FlutterWebRTCBase::MediaTrackForId only local tracks were looked in.
Added also looking in remote tracks.

The problem was visible as described. When joining a session and trying to enable a track inside flutter_webrtc there is a call
`RTCMediaTrack* track = MediaTrackForId(track_id);`

That returned `nullptr` and exception was thrown:
`result->Error("mediaStreamTrackSetEnableFailed", "mediaStreamTrackSetEnable() track is null");`


ashellunts:
cc @hiroshihorie

hiroshihorie:
@cloudwebrtc Can you take a look ? I'm not so familiar for the windows platform..

cloudwebrtc:
LGTM

PR #802, opened 2021-12-07
Fix crash on windows when camera is busy

ashellunts:
Check if capturer is null and do not start video source
in that case.


cloudwebrtc:
Hey, @ashellunts Thank you for the quick fix., Can you use the release version to replace libwebrtc.dll in the PR ? the performance of the debug version of the DLL will be much slower.  The release dll size will be smaller, about 12MB

cloudwebrtc (approved):

ashellunts:
Sure, will do that.

ashellunts:
done

Issue #691, opened 2021-09-20
Listen for frames on media stream


fotiDim:
I would like to perform realtime processing on the frames of the remote video feed using [TFlite](https://pub.dev/packages/tflite). Currently there is no way to do that from what it seems.

I have seen that there is CaptureFrame method but that is a one time thing. I could run that on a timer I suppose but performance might be a problem. Ideally there should be a stream that I can listen on from the Dart side.

arafat1:
Any update on this. It's a good idea to provide converter for platform-specific stream object to/from flutter_webrtc MediaStream object. This will allow anyone to take a flutter_wrbrtc MediaStream, convert it to platform specific stream, perform transformations on it and then convert it back to flutter_webrtc MediaStream. This will enable all sorts of video processing (virtual background, blur, filters...) with minimum code addition to the flutter_webrtc codebase. I can start with web and do a merge request with changes if CLoubdwebrtc plan to go ahead with this. If the Cloudwebrtc team has some different paths/plans in progress, then please share.

cloudwebrtc:
I think the perfect solution is insertable streams, which can export raw data from audio/video track, modify the data and then generate a new track through media_stream_track_generator.
https://github.com/flutter-webrtc/webrtc-interface/tree/feat/insertable-streams/lib/src/insertable_streams

arafat1:
Thanks for sharing the update. This looks very promising, though the current support is only available in chromium based browsers. What would be the raw data format? Byte array (Unit8List)?

fotiDim:
@cloudwebrtc @arafat1 I have been work on a solution for this. I opened a [Draft PR](https://github.com/flutter-webrtc/flutter-webrtc/pull/813) for you to have a look what I have been up to. I am not sure how well it fits with the insertable_streams streams concept you mentioned. Seems it's a different approach. In my attempt I am not exposing frames from this package but instead I allow feeding custom frames into this package.

arafat1:
[...]

PR #813, opened 2021-12-20
[RFC] Add custom video source

fotiDim:
This is a draft PR for now to discuss the approach that I took. This is iOS only for now. There are a couple of things to be sorted out before this PR is considered ready.
The motivation behind it was to allow access on individual video frames and addressing https://github.com/flutter-webrtc/flutter-webrtc/issues/691. Instead of exposing the frames though the FlutterWebRTC package this PR enables feeding custom frames into FlutterWebRTC essentially allowing any kind of preprocessing. Before, only camera and screen capture was possible, now there is also the "custom" option that allows for more flexibility.
The end developer could use the [camera package](https://pub.dev/packages/camera) with `startImageStream((CameraImage image)` which returns all camera frames. Then would be able to do any kind of preprocessing he wants (TFLite, filters etc) and eventually feed the frames into the FlutterWebRTC package.

```
Camera Images -> (Processing w/ TFLite -> Processing w/ Filters -> More Processing) -> broadcastFrame() w/ FlutterWebRTC
````

CI fails because the newly added method needs to also be added in https://github.com/flutter-webrtc/webrtc-interface. I would like to validate my approach before proceeding further. Thoughts?


ycherniavskyi:
@fotiDim why you decided to close the PR?

fotiDim:
> @fotiDim why did you decide to close the PR?

I was expecting for some feedback for too long.

Issue #572, opened 2021-05-29
Web version loopback not work


kadenchoi:
I have deployed the example 
https://rtc-kickapps.web.app

I find the loopback its not work on Web

kadenchoi:
InvalidAccessError: Failed to execute 'setRemoteDescription' on 'RTCPeerConnection': Failed to set remote answer sdp: Failed to apply the description for m= section with mid='0': Answerer must
use either active or passive value for setup attribute.

oofp:
I have the same issue

PR #824, opened 2021-12-29
set setup attribute to active at SDP answer for loopback example

oofp:
Fix error reported at https://github.com/flutter-webrtc/flutter-webrtc/issues/572


cloudwebrtc:
LGTM

PR #825, opened 2021-12-30
feat: support unprocessed audio

cloudwebrtc:
Supports unprocessed audio input, such as midi, singing, musical instruments. or specific sound sources.
Keep the original sound quality as much as possible.


cloudwebrtc:
@kNoAPP hey, you can use this branch for testing (review ready), I added bypassVoiceProcessing for plugin initialize, in the loopback sample you can turn off audio 3A,

````
  final String AUDIO_ECHO_CANCELLATION_CONSTRAINT = 'googEchoCancellation';
  final String AUDIO_AUTO_GAIN_CONTROL_CONSTRAINT = 'googAutoGainControl';
  final String AUDIO_HIGH_PASS_FILTER_CONSTRAINT = 'googHighpassFilter';
  final String AUDIO_NOISE_SUPPRESSION_CONSTRAINT = 'googNoiseSuppression';

  void _makeCall() async {
    final mediaConstraints = <String, dynamic>{
      'audio': {
        'mandatory': {
          AUDIO_ECHO_CANCELLATION_CONSTRAINT: false,
          AUDIO_AUTO_GAIN_CONTROL_CONSTRAINT: false,
          AUDIO_HIGH_PASS_FILTER_CONSTRAINT: false,
          AUDIO_NOISE_SUPPRESSION_CONSTRAINT: false,
        }
      },
      'video': {
        'mandatory': {
          'minWidth':
              '640', // Provide your own width, height and frame rate here
          'minHeight': '480',
          'minFrameRate': '30',
        },
        'facingMode': 'user',
        'optional': [],
      }
    };
````

kNoAPP:
Great news! I will give it a test this week and get back to you.

kNoAPP:
@cloudwebrtc The changes for iOS & Android sound really good! There's a definite improvement to the instrument audio coming through. A few things found in testing:

1. The unprocessed audio mode is not toggleable. Once its set, you can't change it.
2. There's no change on MacOS. The audio is still being filtered here.

If possible, can these be fixed before merging?

cloudwebrtc:
I looked at the audio I/O section of macOS and it seems to be quite different from iOS. still need some time to try to figure it out.

jrmilholland:
@cloudwebrtc - We took over the project from kNoAPP, any update about MacOS?
Also, we ran into an issue on iOS when bypassVoiceProcessing = true, when sending data, the receiver hears their own voice (this doesn't happen when bypassVoiceProcessing = false), any thoughts?
Thanks

cloudwebrtc:
hey, @jrmilholland,  I've had a very busy 2+ months, but I'll have time to work on these issues over the next week.
This may be due to the fact that the device does not handle automatic gain and echo suppression after enabling bypassVoiceProcessing (however, the mic acquisition gain of iOS is very sensitive), so you need to try connecting wired headphones or monitoring headphones to avoid the sound returning to the microphone.

upta:
Hey @cloudwebrtc,

Our client has some additional work he’d like to discuss with you if that’s something you might be interested in.

Would you be open to scheduling an audio/video chat?  We are in the Pacific time zone but could work with whatever works best for you.

I sent you an email as well, but I figured I'd try here as well just in case that lands in a spam filter or something.

Thanks!
Brian

cloudwebrtc:
I'll try to add unprocessed-audio support to macOS this week.

PR #826, opened 2022-01-03
Better handling of `rid` in `sender.parameters.encodings`

hiroshihorie:
- `rid` is missing for `rtpParametersToMap` for both iOS, Android.
- higher priority for updating encoding with same rid. (order of array wouldn't matter)
@davidzhao


davidzhao:
really solid work!

PR #832, opened 2022-01-14
Properly change 'track' property of RTCRtpSender

jjpark-edito:
After following APIs are called, the 'track' property of the RTCRtpSender object was not updated. 

- `RTCRtpSender.setTrack()`
- `RTCRtpSender.replaceTrack()`
- `RTCPeerConnection.removeTrack()`

I fixed to correctly update the 'track' property of RTCRtpSender after processing these APIs.


cloudwebrtc:
LGTM

PR #891, opened 2022-03-24
Fix simulcast factory not sending back EncoderInfo

davidliu:
This fixes Android hardware encoding for H264.

See https://github.com/livekit/client-sdk-android/pull/67 for other discussion.


hiroshihorie (approved):
Very nice

cloudwebrtc:
Very useful fix, I just had the same issue in the flutter_whip plugin, luckily @davidliu  you fixed it.

PR #909, opened 2022-04-11
Bluetooth Permission for Android 12+

BFamzz:
Bluetooth Permission for Android 12 devices and upwards

### DIFF
--- a/android/src/main/java/com/cloudwebrtc/webrtc/utils/RTCBluetoothManager.java
+++ b/android/src/main/java/com/cloudwebrtc/webrtc/utils/RTCBluetoothManager.java
@@ -233,7 +233,8 @@ public class RTCBluetoothManager {
   public void start() {
     ThreadUtils.checkIsOnMainThread();
     Log.d(TAG, "start");
-    if (!hasPermission(apprtcContext, android.Manifest.permission.BLUETOOTH)) {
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/cloudwebrtc/webrtc/utils/RTCBluetoothManager.java
```diff
@@ -233,7 +233,8 @@ public State getState() {
   public void start() {
     ThreadUtils.checkIsOnMainThread();
     Log.d(TAG, "start");
-    if (!hasPermission(apprtcContext, android.Manifest.permission.BLUETOOTH)) {
+    if (!hasPermission(apprtcContext, android.Manifest.permission.BLUETOOTH) 
+        && !hasPermission(apprtcContext, android.Manifest.permission.BLUETOOTH_CONNECT)) {
```
davidliu, line 237: I think this needs to check if on Android 12 first? This value may not exist on lower API versions and crash.

BFamzz: Yes you're right. I'll update. Thanks.

davidliu:
Hey, opened up #913, which includes this fix with version checking, as well as handling requesting the needed permissions. Closing this PR in favor of the other one.

Issue #910, opened 2022-04-12
 type 'int' is not a subtype of type 'String'


hrxiang:
E/flutter (17579): [ERROR:flutter/lib/ui/ui_dart_state.cc(209)] Unhandled Exception: type 'int' is not a subtype of type 'String'
E/flutter (17579): #0      RTCDataChannelNative.eventListener (package:flutter_webrtc/src/native/rtc_data_channel_impl.dart:54:42)
E/flutter (17579): #1      _rootRunUnary (dart:async/zone.dart:1434:47)
E/flutter (17579): #2      _CustomZone.runUnary (dart:async/zone.dart:1335:19)
E/flutter (17579): #3      _CustomZone.runUnaryGuarded (dart:async/zone.dart:1244:7)
E/flutter (17579): #4      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:341:11)
E/flutter (17579): #5      _DelayedData.perform (dart:async/stream_impl.dart:591:14)
E/flutter (17579): #6      _StreamImplEvents.handleNext (dart:async/stream_impl.dart:706:11)
E/flutter (17579): #7      _PendingEvents.schedule.<anonymous closure> (dart:async/stream_impl.dart:663:7)
E/flutter (17579): #8      _rootRun (dart:async/zone.dart:1418:47)
E/flutter (17579): #9      _CustomZone.run (dart:async/zone.dart:1328:19)
E/flutter (17579): #10     _CustomZone.runGuarded (dart:async/zone.dart:1236:7)
E/flutter (17579): #11     _CustomZone.bindCallbackGuarded.<anonymous closure> (dart:async/zone.dart:1276:23)
E/flutter (17579): #12     _rootRun (dart:async/zone.dart:1426:13)
E/flutter (17579): #13     _CustomZone.run (dart:async/zone.dart:1328:19)
E/flutter (17579): #14     _CustomZone.runGuarded (dart:async/zone.dart:1236:7)
E/flutter (17579): #15     _CustomZone.bindCallbackGuarded.<anonymous closure> (dart:async/zone.dart:1276:23)
E/flutter (17579): #16     _microtaskLoop (dart:async/schedule_microtask.dart:40:21)
E/flutter (17579): #17     _startMicrotaskLoop (dart:async/schedule_microtask.dart:49:5)

flutter_webrtc: ^0.8.5
flutter:  Channel stable  2.10.3

simonbengtsson:
[...]

PR #912, opened 2022-04-13
fix : type 'int' is not a subtype of type 'String' #910

zjzhang-cn:


lib/src/native/rtc_data_channel_impl.dart
```diff
@@ -51,7 +51,7 @@ class RTCDataChannelNative extends RTCDataChannel {
     final Map<dynamic, dynamic> map = event;
     switch (map['event']) {
       case 'dataChannelStateChanged':
-        _dataChannelId = int.tryParse(map['id']);
+        _dataChannelId = int.tryParse(map['id'].toString());
```
ycherniavskyi, line 54: But `tryParse` could return `null`, so it seems that there must be some additional condition about that.

cloudwebrtc: I'm not sure if it's a platform-specific exception, the type of id should be enforced at the native layer, I think we have to fix it at the native layer

cloudwebrtc: I mean, the `id` must exist in `dataChannelStateChanged` or other `events`, and the type must be `int`, so as to avoid exceptions,

PR #929, opened 2022-05-01
Support null tracks in sender replace track and set track

itsthetaste:
I've added an example page as well showing starting and stopping the local cam and mic before the remote connection is established.
Relies upon a PR #4 in webrtc-interface
https://github.com/flutter-webrtc/webrtc-interface/pull/4


cloudwebrtc:
hey @itsthetaste, I'll merge after testing all this week, thanks for your contribution

cloudwebrtc (approved):

PR #945, opened 2022-05-11
Update build.gradle - switch from jCenter to mavenCentral

RobotKSR:
Switch to mavenCentral() in build.gradle because "jcenter()" (jcenter.bintray.com) servers are down from now.


ycherniavskyi:
Just more extended explanation of necessity of this switch - https://stackoverflow.com/a/50726436/9967488

ycherniavskyi (commented):
Finished.

PR #952, opened 2022-05-15
fix: "Reply already submitted" error and setVolume() not working on remote streams

GeorgeAmgad:
Previously if something unexpected happened the flow calls `resultError()` then calls `result.success()` which throws "Reply already submitted" because result is already called inside `resultError()`.

### DIFF
--- a/android/src/main/java/com/cloudwebrtc/webrtc/MethodCallHandlerImpl.java
+++ b/android/src/main/java/com/cloudwebrtc/webrtc/MethodCallHandlerImpl.java
@@ -545,7 +545,7 @@ public class MethodCallHandlerImpl implements MethodCallHandler, StateProvider {
           params.putString("type", sdp.type.canonicalForm());
           result.success(params.toMap());
         } else {
-          resultError("getLocalDescription", "peerConnection is nulll", result);
+          resultError("getLocalDescription", "peerConnection is null", result);
         }
         break;
       }
@@ -563,7 +563,7 @@ public class MethodCallHandlerImpl implements MethodCallHandler, StateProvider {
             result.success(params.toMap());
           }
         } else {
-          resultError("getRemoteDescription", "peerConnection is nulll", result);
+          resultError("getRemoteDescription", "peerConnection is null", result);
         }
         break;
       }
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/cloudwebrtc/webrtc/MethodCallHandlerImpl.java
```diff
@@ -1206,9 +1206,11 @@ public void mediaStreamAddTrack(final String streamId, final String trackId, Res
         }
       } else {
         resultError("mediaStreamAddTrack", "mediaStreamAddTrack() track [" + trackId + "] is null", result);
+        return;
       }
     } else {
       resultError("mediaStreamAddTrack", "mediaStreamAddTrack() stream [" + streamId + "] is null", result);
+      return;
     }
     result.success(null);
```
ycherniavskyi, line 1215: I propose not to add `return`, but to move `result.success(null);` in the proper place. The flow will be much more readable.

GeorgeAmgad:
I also added a tested fix for `setVolume()` not triggering for remote streams on Android.

android/src/main/java/com/cloudwebrtc/webrtc/MethodCallHandlerImpl.java
```diff
@@ -1201,16 +1201,17 @@ public void mediaStreamAddTrack(final String streamId, final String trackId, Res
       if (track != null) {
         if (track.kind().equals("audio")) {
           mediaStream.addTrack((AudioTrack) track);
+          result.success(null);
         } else if (track.kind().equals("video")) {
           mediaStream.addTrack((VideoTrack) track);
+          result.success(null);
         }
       } else {
```
ycherniavskyi, line 1209: To provide result in all possible flow cases I propose to add:
```java
String kind = track.kind();
...
resultError("mediaStreamRemoveTrack", "mediaStreamAddTrack() track [" + trackId + "] has unsupported type: " + kind, result);
```
between this lines.

ycherniavskyi: Or maybe move `result.success(null);` to that place. But as for me error will more relevant.

GeorgeAmgad: Makes sense, I added the error message.

ycherniavskyi (approved):

ycherniavskyi:
@GeorgeAmgad I am going to merge this PR for now, but in future please do not mix more then one fix/improvement in one PR ("replay already submitted" and "setVolume() not working on remote" in this case).

Issue #948, opened 2022-05-13
MacOS libyuv_internal.a needs to be built for arm64 support


jbrisko:
With the release of Flutter 3, and support for Universal binaries we need a version of [libyuv_internal.a](https://github.com/flutter-webrtc/flutter-webrtc/blob/master/macos/third_party/lib/libyuv_internal.a) that is compiled for universal binary support. Currently, it cannot link to the library due to missing symbols.

**To Reproduce**
Build an app with arm64 support on Mac.

**Expected behavior**
It should build

**Platform information**
Flutter 3, Mac OS 12.3.1

letungcntt:
``` console
ld: warning: ignoring file /Users/tungle/.pub-cache/git/flutter-webrtc/macos/third_party/lib/libyuv_internal.a, building for macOS-arm64 but attempting to link with file built for macOS-x86_64
Undefined symbols for architecture arm64:
  "_I420Rotate", referenced from:
      -[FlutterRTCVideoRenderer correctRotation:withRotation:] in FlutterRTCVideoRenderer.o
  "_I420ToARGB", referenced from:
      -[FlutterRTCVideoRenderer copyI420ToCVPixelBuffer:withFrame:] in FlutterRTCVideoRenderer.o
  "_I420ToBGRA", referenced from:
      -[FlutterRTCVideoRenderer copyI420ToCVPixelBuffer:withFrame:] in FlutterRTCVideoRenderer.o
  "_I420ToNV12", referenced from:
      -[FlutterRTCVideoRenderer copyI420ToCVPixelBuffer:withFrame:] in FlutterRTCVideoRenderer.o
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
** BUILD FAILED **

leesc81:
I have the same problem.
At build time, the following log appears:

Launching lib/main.dart on macOS in debug mode...
Building macOS application...
objc[95601]: Class AppleTypeCRetimerRestoreInfoHelper is implemented in both /usr/lib/libauthinstall.dylib (0x1e1499eb0) and /Library/Apple/System/Library/PrivateFrameworks/MobileDevice.framework/Versions/A/MobileDevice (0x1040f04f8). One of the two will be used. Which one is undefined.
[...]

PR #953, opened 2022-05-16
Flutter 3.0 crash fixes

hiroshihorie:
setStreamHandler needs to be called on `main` thread.

Calling setStreamHandler from a WebRTC delegate (which is invoked on it's internal thread) causes a thread check failure and crash.

### DIFF
--- a/common/darwin/Classes/FlutterRTCPeerConnection.m
+++ b/common/darwin/Classes/FlutterRTCPeerConnection.m
@@ -504,8 +504,11 @@
 
     dataChannel.eventChannel = eventChannel;
     dataChannel.flutterChannelId = dataChannelId;
-    [eventChannel setStreamHandler:dataChannel];
 
+    dispatch_async(dispatch_get_main_queue(), ^{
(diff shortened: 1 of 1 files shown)


hiroshihorie:
Digging into this, i see how `peerConnection.dataChannels` dictionary is being manipulated in the WebRTC's delegate 
which doesn't look like a good idea. At some point if there is simultaneous access from main thread and WebRTC's thread it will crash.

davidzhao (approved):
got it. so this will be a short term fix until we correct the manipulation in callback?

hiroshihorie:
This fix is fine, but I see other potential danger like accessing `peerConnection.dataChannels` within WebRTC's thread which is also accessed from `main` 🙂

cloudwebrtc:
hi @hiroshihorie https://github.com/flutter-webrtc/flutter-webrtc/issues/948
Here is another question about flutter 3.0, should we solve it together before releasing a new version?

hiroshihorie:
@cloudwebrtc Oh, we probably should... nice PR !

Issue #782, opened 2021-11-23
About libyuv


yangfanyu:
https://github.com/flutter-webrtc/flutter-webrtc/issues/168

https://github.com/flutter-webrtc/flutter-webrtc/pull/188

conflict with

https://github.com/flutter-webrtc/flutter-webrtc-demo/issues/76

https://github.com/flutter-webrtc/flutter-webrtc-demo/issues/76#issuecomment-976444789

I am not familiar with ios, may be  change `#include "libyuv/xxx.h"` to `#include "Libyuv/xxx.h" (case sensitive)  in  libyuv library can resolved them?

mrcsh:
I am seeing link errors with libyuv with an M1 mac which I know isn't 100% there yet, at least when compiling for the emulator.
I can build to an actual device and archive from the UI works but command line archive stopped working for some reason.

PR #954, opened 2022-05-17
fix: Use RTCYUVHelper instead of external libyuv library.

cloudwebrtc:
fix https://github.com/flutter-webrtc/flutter-webrtc/issues/948
fix https://github.com/flutter-webrtc/flutter-webrtc/issues/782

Tested on macOS (intel) and iPhone 13. Works fine.


hiroshihorie:
@cloudwebrtc Hello Duan, I know you are working on this but do you think we can merge this and have a new release tonight ? Thanks a lot !

PR #972, opened 2022-06-02
Implemented features for android

abhay-s-rawat:
Immplemented for android:
created pull request for webrtc_interface 1.0.5

onBufferedAmountChange
bufferedAmount
onBufferedAmountLow
bufferedAmountLowThreshold

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 [0.8.8] - 2022-05-31
 
 * [Android] Added onBufferedAmountChange callback which will return currentBuffer and changedBuffer and implemented bufferedAmount.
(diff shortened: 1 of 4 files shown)


cloudwebrtc (approved):

cloudwebrtc:
Nice patch, we need to make iOS/Windows support the same before uploading to `pub.dev`.

abhay-s-rawat:
Yes but I need to take a look at ios other platform code and how it is working which might take time because I am not familiar with swift. It was minor update for android because it uses java and I used that a long time ago that's why I was familiar with that code.

abhay-s-rawat:
Can we push changes for android first in git then after adding those features on other platform you can push to pub.dev.

cloudwebrtc:
sure, no problem

PR #976, opened 2022-06-05
Fixes ordering issue with sendEncodings

digitalix:


common/darwin/Classes/FlutterRTCPeerConnection.m
```diff
@@ -506,19 +506,18 @@ - (void)peerConnection:(RTCPeerConnection*)peerConnection didOpenDataChannel:(RT
...
        [eventChannel setStreamHandler:dataChannel];
+       FlutterEventSink eventSink = peerConnection.eventSink;
+       if(eventSink){
+           eventSink(@{
+                       @"event" : @"didOpenDataChannel",
+                       @"id": dataChannelId,
+                       @"label": dataChannel.label,
+                       @"flutterId": flutterChannelId
+                       });
+       }
```
ycherniavskyi, line 520: What is the reason to move `eventSink` call into `dispatch_async`?

ycherniavskyi (commented):
It will be much easier to review and merge if you provide at last some general description (possible with links to spec/doc) of your pull request.

common/darwin/Classes/FlutterRTCPeerConnection.m
digitalix: Oh sorry this wasn't supposed to be here. It's from another pull request https://github.com/flutter-webrtc/flutter-webrtc/pull/975.

Should I remove this commit from here?

ycherniavskyi: Yes, it will be great. Mixing different pull requests in one is not good.

digitalix: Ok, removed!

ycherniavskyi (commented):
I just get dipper in history of this lines and found that your commit revert the  40bde2ae3db08691427cb474db48618754256083.
That is why for sure we need to add @cloudwebrtc in this discussion.

Issue #989, opened 2022-07-07
Crash when call findLongInt


xiaowei-guan:
```cpp
 inline int64_t findLongInt(const EncodableMap& map, const std::string& key) {
  for (auto it : map) {
    if (key == GetValue<std::string>(it.first) &&
        (TypeIs<int64_t>(it.second) || TypeIs<int32_t>(it.second)))
      return GetValue<int64_t>(it.second);
  }

  return -1;
}
 ```
 If we set a value by int32_t but we get it by type int64_t, will throw a std::bad_variant_access exception.
 
we can replace findLongInt with LongValue() which bleong to encodable_value.h

PR #990, opened 2022-07-07
Fix variant exception of findLongInt.

wanchao-xu:
If the EncodableValue is int32_t, findLongInt will throw a std::bad_variant_access exception.
Please refer to https://github.com/flutter-webrtc/flutter-webrtc/issues/989

### DIFF
--- a/common/cpp/include/flutter_webrtc_base.h
+++ b/common/cpp/include/flutter_webrtc_base.h
@@ -86,9 +86,13 @@ inline int findInt(const EncodableMap& map, const std::string& key) {
 
 inline int64_t findLongInt(const EncodableMap& map, const std::string& key) {
   for (auto it : map) {
-    if (key == GetValue<std::string>(it.first) &&
-        (TypeIs<int64_t>(it.second) || TypeIs<int32_t>(it.second)))
-      return GetValue<int64_t>(it.second);
+    if (key == GetValue<std::string>(it.first)) {
+      if (TypeIs<int64_t>(it.second)) {
+        return GetValue<int64_t>(it.second);
+      } else if (TypeIs<int32_t>(it.second)) {
+        return GetValue<int32_t>(it.second);
+      }
+    }
   }
 
   return -1;


cloudwebrtc:
hey @wanchao-xu, thanks for your fixes.

cloudwebrtc (approved):

ycherniavskyi:
Could you please give more context when the actual value could be `int32_t`?

I just looked into the code and see only two ([1](https://github.com/flutter-webrtc/flutter-webrtc/blob/1436ac730c7b6de8e928cb4da7535a4bec8b8e5f/common/cpp/src/flutter_webrtc.cc#L459) and [2](https://github.com/flutter-webrtc/flutter-webrtc/blob/1436ac730c7b6de8e928cb4da7535a4bec8b8e5f/common/cpp/src/flutter_webrtc.cc#L470)) places where this `findLongInt` method used. And in all places, it used to get `textureId`.

From other side only one place where `textureId` [set](https://github.com/flutter-webrtc/flutter-webrtc/blob/1436ac730c7b6de8e928cb4da7535a4bec8b8e5f/common/cpp/src/flutter_video_renderer.cc#L141) and it always set with [`int64_t` type](https://github.com/flutter-webrtc/flutter-webrtc/blob/1436ac730c7b6de8e928cb4da7535a4bec8b8e5f/common/cpp/src/flutter_video_renderer.cc#L138).

So I'm just curious where this exception comes from 🤔. And what if instead of extending `findLongInt` method to convert `int32_t` to `int64_t`, we need to add a separate method to return `int32_t`.

wanchao-xu:
I'm in charging of flutter-tizen(https://github.com/flutter-tizen). Now I support flutter-webrtc on tizen platform. And the `textureId` got from tizen TV is `int32_t`.

xiaowei-guan:
````cpp
  int64_t LongValue() const {
    if (std::holds_alternative<int32_t>(*this)) {
      return std::get<int32_t>(*this);
    }
    return std::get<int64_t>(*this);
  }
````
Actually, we can try use this method.
https://github.com/flutter/engine/blob/main/shell/platform/common/client_wrapper/include/flutter/encodable_value.h#L212

cloudwebrtc:
> I'm in charging of flutter-tizen(https://github.com/flutter-tizen). Now I support flutter-webrtc on tizen platform. And the `textureId` got from tizen TV is `int32_t`.

@wanchao-xu cool, It looks like flutter-webrtc supports a new platform, glad heard the news

cloudwebrtc:
hey @wanchao-xu, I'm a little curious, is there any calling difference between the flutter-webrtc CPP plugin on the flutter tizen platform? Or only a few differences.

BTW, I'd love to accept PRs to support flutter-tizen if you like. This will make flutter-webrtc even better.

PR #992, opened 2022-07-07
Add screen sharing event listener.

cloudwebrtc:


common/darwin/Classes/FlutterRTCDesktopCapturer.m
```diff
@@ -277,25 +301,149 @@ -(RTCDesktopSource *)getSourceById:(NSString *)sourceId {
...
+    if(_eventSink) {
+        _eventSink(@{
+            @"event": @"desktopSourceNameChanged",
+            @"id": source.sourceId,
+            @"name": source.name,
+        });
+    }
+}
+
+- (void)didDesktopSourceThumbnailChanged:(RTC_OBJC_TYPE(RTCDesktopSource) *) source {
```
davidzhao, line 415: oh this is great. would help us utilize these events in a consistent way.

macos/flutter_webrtc.podspec
```diff
@@ -3,7 +3,7 @@
 #
 Pod::Spec.new do |s|
   s.name             = 'flutter_webrtc'
-  s.version          = '0.7.1'
+  s.version          = '0.8.11'
```
davidzhao, line 6: that's a large jump. why `0.8.11` instead of `0.8.0`?

davidzhao (approved):

macos/flutter_webrtc.podspec
cloudwebrtc: I want to make its version consistent with flutter-webrtc, but it seems difficult to align the two, so you are right, just jump to 0.8.0.

davidzhao: oh got it.. making them aligned makes a lot of sense!

common/darwin/Classes/FlutterRTCDesktopCapturer.h
```diff
@@ -16,6 +16,13 @@
 -(void)getDesktopSources:(NSDictionary *)argsMap
              result:(FlutterResult)result;
 
+-(void)stopDesktopSourcesRefersh:(NSDictionary *)argsMap
```
davidliu, line 19: typo:

```suggestion
-(void)stopDesktopSourcesRefresh:(NSDictionary *)argsMap
```

common/darwin/Classes/FlutterRTCDesktopCapturer.m
```diff
@@ -277,25 +301,149 @@ -(RTCDesktopSource *)getSourceById:(NSString *)sourceId {
...
+            @"thumbnail": data 
+        });
+    }
+}
+
+#pragma mark - RTCDesktopCapturerDelegate delegate
+
+-(void)didSourceCaptureStart:(RTCDesktopCapturer *) capturer {
+    NSLog(@"didSourceCaptureStart");
+    _interruptSourcesRefresh = YES;
```
davidliu, line 432: Can you explain why this is needed? Does source capturing interfere with the updating of the sources list?

common/darwin/Classes/FlutterWebRTCPlugin.m
```diff
@@ -969,7 +972,10 @@ - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult) result
     } else  if ([@"getDesktopSources" isEqualToString:call.method]){
         NSDictionary* argsMap = call.arguments;
         [self getDesktopSources:argsMap result:result];
-    }  else  if ([@"getDesktopSourceThumbnail" isEqualToString:call.method]){
+    } else  if ([@"stopDesktopSourcesRefersh" isEqualToString:call.method]) {
```
davidliu, line 975: typo:

```suggestion
    } else  if ([@"stopDesktopSourcesRefresh" isEqualToString:call.method]) {
```

lib/src/native/desktop_capturer_impl.dart
```diff
@@ -64,15 +131,16 @@ class DesktopCapturerNative extends DesktopCapturer {
...
+      desktopSource.thumbnail = source['thumbnail'] as Uint8List;
+      _sources[desktopSource.id] = desktopSource;
     }
-    return sources;
+    */
+    return _sources.values.toList();
+  }
+
+  @override
+  Future<void> stopRefershSources() async {
```
davidliu, line 142:
```suggestion
  Future<void> stopRefreshSources() async {
```

lib/src/native/desktop_capturer_impl.dart
```diff
@@ -64,15 +131,16 @@ class DesktopCapturerNative extends DesktopCapturer {
...
+    await WebRTC.invokeMethod('stopDesktopSourcesRefersh');
```
davidliu, line 143:
```suggestion
    await WebRTC.invokeMethod('stopDesktopSourcesRefresh');
```

davidliu:
There's a bunch of `refersh` typos, may want to just find and replace all of them.

common/darwin/Classes/FlutterRTCDesktopCapturer.m
cloudwebrtc: Actually I'm going to delete the use of timer in native to refresh the list tomorrow (only `getSources`, `updateSources` two methods), so these codes will be removed

cloudwebrtc:
After modification, use `getSources` to generate a new list, and then call `updateSources` (such as the selection dialog is opening) to refresh the preview image, add or remove window sources, so just refresh with timer in the dart code.

PR #1015, opened 2022-07-19
Feat/screen capture event listener for mac/win

cloudwebrtc:
Consistent with macOS API calls, detailed testing is required. after this feature is merged, I think it is possible to bump the version to 0.9.0

5b6d30742e40f8c5e16444b840a4abb
9f4e7d866fb980f2ef821c208bcb6cd

### DIFF
--- a/common/cpp/src/flutter_screen_capture.cc
+++ b/common/cpp/src/flutter_screen_capture.cc
@@ -160,14 +160,24 @@ void FlutterScreenCapture::OnMediaSourceThumbnailChanged(
 }
 
 void FlutterScreenCapture::OnStart(scoped_refptr<RTCDesktopCapturer> capturer) {
+  std::cout << " OnStart: " << capturer->source()->id().std_string()
+            << std::endl;
 }
 
 void FlutterScreenCapture::OnPaused(
-    scoped_refptr<RTCDesktopCapturer> capturer) {}
+    scoped_refptr<RTCDesktopCapturer> capturer) {
+  std::cout << " OnPaused: " << capturer->source()->id().std_string()
+            << std::endl;
+}
 
-void FlutterScreenCapture::OnStop(scoped_refptr<RTCDesktopCapturer> capturer) {}
+void FlutterScreenCapture::OnStop(scoped_refptr<RTCDesktopCapturer> capturer) {
+  std::cout << " OnStop: " << capturer->source()->id().std_string()
+            << std::endl;
+}
 
 void FlutterScreenCapture::OnError(scoped_refptr<RTCDesktopCapturer> capturer) {
+  std::cout << " OnError: " << capturer->source()->id().std_string()
+            << std::endl;
 }
 
 void FlutterScreenCapture::GetDesktopSourceThumbnail(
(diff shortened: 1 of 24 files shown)


davidzhao:
looking good! but will let either @hiroshihorie or @davidliu to review.

just noticed a typo here in the screen shot: `Entrire` -> `Entire`

cloudwebrtc:
> just noticed a typo here in the screen shot: Entrire -> Entire

ha, thanks,  fixed now.

common/cpp/src/flutter_screen_capture.cc
```diff
@@ -8,86 +8,213 @@
...
+  for (size_t i = 0; i < size; i++) {
+    std::string type = GetValue<std::string>(types[i]);
+    if (type == "window") {
+      scoped_refptr<RTCDesktopMediaList> window;
+      auto it = medialist_.find(kWindow);
+      if (it != medialist_.end()) {
+        window = (*it).second;
+      } else {
+        window = base_->desktop_device_->GetDesktopMediaList(kWindow);
+        window->RegisterMediaListObserver(this);
```
davidliu, line 47: Do these need to be deregistered when `sources_` is cleared?

common/cpp/src/flutter_screen_capture.cc
```diff
@@ -8,86 +8,213 @@
...
-    info[EncodableValue("type")] = EncodableValue(source.type);
-    list.push_back(EncodableValue(info));
+void FlutterScreenCapture::GetDesktopSources(const EncodableList &types, std::unique_ptr<MethodResult<EncodableValue>> result) {
+  EncodableList sources;
+  sources_.clear();
+  size_t size = types.size();
+  for (size_t i = 0; i < size; i++) {
+    std::string type = GetValue<std::string>(types[i]);
+    if (type == "window") {
+      scoped_refptr<RTCDesktopMediaList> window;
```
davidliu, line 41: Can we refactor this into a separate method? We seem to reuse this code elsewhere, and it's duplicated below for screen type (with the type seemingly the only difference).

common/cpp/src/flutter_screen_capture.cc
```diff
@@ -8,86 +8,213 @@
...
+      auto it = medialist_.find(kWindow);
+      if (it != medialist_.end()) {
+        window = (*it).second;
+      } else {
+        window = base_->desktop_device_->GetDesktopMediaList(kWindow);
+        window->RegisterMediaListObserver(this);
+        medialist_[kWindow] = window;
+      }
+      window->UpdateSourceList(true);
```
davidliu, line 50: We pass in `true` here, but everywhere we pass in false? Is there something specific to this particular code path?

cloudwebrtc: Here is the method in libwebrtc. When true, it will rebuild the entire source list. When it is false, it will use add or remove to refresh the list. So I will give it false when updateDesktopSources
```cpp
virtual int32_t UpdateSourceList(bool force_reload = false, bool get_thumbnail = true) = 0;
```

common/cpp/src/flutter_screen_capture.cc
cloudwebrtc: yup, it makes sense, I will refactor them to make the code more concise.

common/cpp/src/flutter_screen_capture.cc
davidliu: So if it's false, the `OnMediaSourceAdded/Removed` callbacks will get called, while if it's true, those callbacks won't be called?

davidliu (approved):
LGTM

common/cpp/src/flutter_screen_capture.cc
cloudwebrtc: If it is true, all items will be removed and OnMediaSourceRemoved will be called, when false, only the window sources that are invalidated or closed will be removed

PR #1048, opened 2022-08-10
Update MethodCallHandlerImpl.java - Fix renderer null object

duyjack:
Fix get render null because renders is map, should use valueAt for avoiding null when get item

Explain:
when RTCVideoRenderer initialize, entry.id() is increase 1
if i create 4 RTCVideoRenderer and call initialize, i have 0 1 2 3
I delete 3 4, i have 0 1
I recreate 2 RTCVideoRenderer, i will have 0 1 4 5
So when use renders.get(i) with i = 2 or 3, renderer is null


cloudwebrtc:
hey @duyjack, thanks for finding this bug.

cloudwebrtc (approved):

PR #1067, opened 2022-08-31
Avoid renderer initialization multiple times

cloudwebrtc:


wer-mathurin (commented):
As a general comment, there is multiple place that we are calling await without a try catch.  So there is are potential bugs, but this is not the intent of this fix.

Other than that, LGTM

If you want to resolve the await in another PR.  Just add a comment

PR #1068, opened 2022-08-31
Added callback onFirstFrameRendered

MohamedRisaldarTA:
This PR will provide a callback to notify user that the first frame is rendered.
User can use it to know when first frame of a stream is started playing.


MohamedRisaldarTA:
Need to wait until https://github.com/flutter-webrtc/webrtc-interface/pull/10 is merged

PR #1091, opened 2022-09-25
fix: Make sure local stream/track dispose correctly.

cloudwebrtc:
fix https://github.com/livekit/client-sdk-flutter/issues/166


davidzhao (approved):
lgtm

PR #1103, opened 2022-10-02
Dispose streams and connections

mouEsam:
Premise: When the app is swiped off on android, `dispose` functions are not called in flutter as the engine is being disposed. Thus the streams, tracks, as well as the peer connections leak and await being deallocated by the system upon memory purging of the process. Until that happens the mic and phone activity indicators on modern android versions show that they are being used by the "supposedly" closed app. 

This PR closes and disposes of the local streams, tracks, connections when the engine is detached instead of just clearing the lists. As this already leaks the undisposed connections.


cloudwebrtc:
lgtm

PR #1123, opened 2022-10-14
Have same remote streams id then found wrong MediaStream.

chifandeyu:
In multiple peerconnection find stream id by for loop. But stream id for each peerconnection may be the same.


cloudwebrtc (approved):
lgtm

PR #1132, opened 2022-10-18
fix compile for linux

doesnotexist:
For some reason, clang fails to recognize that EncodableList and EncodableMap as acceptable variant types for EncodableValue


cloudwebrtc:
hey @doesnotexist, Thanks for your contribution, there is currently no linux plugin support in main branch, can you try pushing it to [feat/support-linux](https://github.com/flutter-webrtc/flutter-webrtc/tree/feat/support-linux) branch?
I did some work in this branch

PR #1162, opened 2022-11-11
fix: Fixed the bug that the mic indicator light was still on when mic recording was stopped

cloudwebrtc:


common/darwin/Classes/AudioUtils.m
```diff
@@ -23,7 +23,7 @@ + (void)ensureAudioSessionWithRecording:(BOOL)recording {
                        error:nil];
         [session setMode:config.mode error:nil];
         [session unlockForConfiguration];
-    } else if (!recording && (session.category == AVAudioSessionCategoryAmbient
+    } else if (!recording || (session.category == AVAudioSessionCategoryAmbient
```
davidzhao, line 26: would this have the side effect of overriding a user's custom session category? i.e. what if I wanted `playbackAndRecord`, but then it kept overriding it to `playback`? 

before we would only override when we had to.

also, would it cause audio glitch if one moves from playbackAndRecord to playback on BT?

cloudwebrtc: Side effects can be avoided by not modifying `config.categoryOptions`.

```patch
diff --git a/common/darwin/Classes/AudioUtils.m b/common/darwin/Classes/AudioUtils.m
index c5495af..3dcd4de 100644
--- a/common/darwin/Classes/AudioUtils.m
+++ b/common/darwin/Classes/AudioUtils.m
@@ -26,7 +26,6 @@ + (void)ensureAudioSessionWithRecording:(BOOL)recording {
     } else if (!recording || (session.category == AVAudioSessionCategoryAmbient
                               || session.category == AVAudioSessionCategorySoloAmbient)) {
         config.category = AVAudioSessionCategoryPlayback;
-        config.categoryOptions = 0;
         config.mode = AVAudioSessionModeDefault;
         
         // upgrade from ambient if needed
```

cloudwebrtc: > also, would it cause audio glitch if one moves from playbackAndRecord to playback on BT?

Tested with airpods, there is no audio glitch after switching

davidzhao: ah ha, that's great. thanks for confirming.

davidzhao (approved):

PR #1189, opened 2022-12-02
Media recorder support for iOS/OSX

rostopira:


rostopira:
Change to interface should be merged https://github.com/flutter-webrtc/webrtc-interface/pull/13

rostopira:
Well, alternative solution for recording audio should be possible.
But only if we expose `setAudioDeviceModule:` method of `RTCPeerConnectionFactoryBuilder` ([src here](https://webrtc.googlesource.com/src/+/refs/heads/main/sdk/objc/api/peerconnection/RTCPeerConnectionFactoryBuilder.h#42))
But it uses C++ interface.
In other words, some changes would be required in WebRTC sdk.
Would be cool if we make some ability to apply patches for webrtc before build.

rostopira:
Nevermind, I've found where you apply [patches](https://github.com/webrtc-sdk/webrtc-build/tree/main/build/patches)
I will try to create patch later

bantunes22:
Can we have this merged into main?

PR #1190, opened 2022-12-02
Handle Platform Exception on addCandidate

MohamedRisaldarTA:
iOS throws 
[VERBOSE-2:ui_dart_state.cc(198)] Unhandled Exception: PlatformException(AddIceCandidateFailed, Error The remote description was null, null, null) sometimes.
and this change will help to handle it.


cloudwebrtc (approved):
lgtm

PR #1202, opened 2022-12-12
Keep the audio switch after stopping

mouEsam:
The audio switch is nullified when stopping which disallows further use after a single stop.
Keeping the reference doesn't cause any memory leaks as the encapsulated fields are removed on stop.


cloudwebrtc (approved):
lgtm

PR #1206, opened 2022-12-15
Enhance RTC video view with placeholder builder property

ycherniavskyi:
This tiny enhancement of `RTCVideoView` allows simplifying add meaningful placeholder widget while `RTCVideoRenderer` is preparing/initialing.

Also, I am going to use it within integration tests that automate app screenshots creation.

### DIFF
--- a/lib/src/native/rtc_video_view_impl.dart
+++ b/lib/src/native/rtc_video_view_impl.dart
@@ -55,13 +55,12 @@ class RTCVideoView extends StatelessWidget {
               child: Transform(
                 transform: Matrix4.identity()..rotateY(mirror ? -pi : 0.0),
                 alignment: FractionalOffset.center,
-                child:
-                    videoRenderer.textureId != null && videoRenderer.renderVideo
-                        ? Texture(
-                            textureId: videoRenderer.textureId!,
-                            filterQuality: filterQuality,
-                          )
-                        : placeholderBuilder?.call(context) ?? Container(),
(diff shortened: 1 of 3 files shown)


lib/src/native/rtc_video_view_impl.dart
```diff
@@ -53,13 +55,13 @@ class RTCVideoView extends StatelessWidget {
...
                 alignment: FractionalOffset.center,
-                child: videoRenderer.textureId != null &&
-                        videoRenderer.srcObject != null
-                    ? Texture(
-                        textureId: videoRenderer.textureId!,
-                        filterQuality: filterQuality,
-                      )
-                    : Container(),
+                child:
+                    videoRenderer.textureId != null && videoRenderer.renderVideo
```
wer-mathurin, line 59: @ycherniavskyi Good idea to use the renderVideo bool to perform the check.  After checking the implementation of the RTCVideoRenderer of the Native and Web, I think we can safely move the  video.textureId != null under the umbrella of renderVideo to make the code more readable for the widget.  

`bool get renderVideo => srcObject != null && _textureId != null;`

Is this sounds resonable?

ycherniavskyi: I agree. Also, I wouldn't say I like this extra condition in the native widget.

wer-mathurin (approved):

wer-mathurin:
@cloudwebrtc and @ycherniavskyi: LGTM

Issue #1212, opened 2022-12-21
Linux fatal error when building


krillefear:
**Describe the bug**

When building on Linux I get this error:

```sh
[  +96 ms] Building Linux application... (completed in 6.1s)
[  +26 ms] (fluffychat:230187): Gtk-WARNING **: 20:33:18.576: Locale not supported by C library.
[        ]      Using the fallback 'C' locale.
[ +131 ms] #
[        ] # Fatal error in: ../../media/engine/adm_helpers.cc, line 39
[        ] # last system error: 0
[        ] # Check failed: 0 == adm->Init() (0 vs. -1)
[  +84 ms] # Failed to initialize the ADM.
[   +2 ms] Error waiting for a debug connection: The log reader stopped unexpectedly, or never started.
[   +1 ms] Error launching application on Linux.
```

**To Reproduce**

Build on Ubuntu 22.10 with Flutter 3.3.9

**Expected behavior**

Just builds

**Platform information**

* **Flutter version**: 3.3.9
* **Plugin version**: 0.9.18
* **OS**:  Ubuntu
* **OS version**: 22.10

cloudwebrtc:
The installation of snap will cause the libwebrtc link of flutter-webrtc to fail due to the inconsistent glibc version. You can try to manually install the flutter sdk, which can compile and run normally.

https://docs.flutter.dev/get-started/install/linux#install-flutter-manually

krillefear:
No, I'm using the Flutter version from flutter.dev directly

cloudwebrtc:
Can you try running `flutter run -d linux -v` and observe the log output?

cloudwebrtc:
From the log, it seems that the runtime failed, which is related to the failure to initialize the audio device module.
```
[        ] # Fatal error in: ../../media/engine/adm_helpers.cc, line 39
[        ] # last system error: 0
[        ] # Check failed: 0 == adm->Init() (0 vs. -1)
[  +84 ms] # Failed to initialize the ADM.
```

krillefear:
> Can you try running `flutter run -d linux -v` and observe the log output?



```sh
[        ] -- Installing: /home/krille/HADTF/Flutter/fluffychat/build/linux/x64/debug/bundle/data/flutter_assets/shaders
[        ] -- Installing: /home/krille/HADTF/Flutter/fluffychat/build/linux/x64/debug/bundle/data/flutter_assets/shaders/ink_sparkle.frag
[...]

PR #1214, opened 2022-12-23
chore: Disable linux support until it is fixed

krille-chan:
Until https://github.com/flutter-webrtc/flutter-webrtc/issues/1212 is fixed


krillefear:
Little update: I still cannot build my app with latest main on Linux and need to have Linux disabled to have a working solution

PR #1217, opened 2023-01-02
Add a method to check if speakerphone is enabled or not

moham96:


lib/src/helper.dart
```diff
@@ -59,6 +59,12 @@ class Helper {
     );
   }
 
+  /// Check if speakerPhone is enabled or not for Flutter native.
+  /// for iOS/Android only
+  static Future<bool> isSpeakerPhoneEnabled() async {
+    return WebRTC.invokeMethod('isSpeakerPhoneEnabled').then((value) => value);
```
ourdex, line 65: I see that implementation is only for Android. What about for iOS?

Issue #1129, opened 2022-10-17
[Android] Crash when try to call the captureFrame function on a VideoTrack


alexlwc11:
**Describe the bug**
Every time I try to call the captureFrame on a video track, the app will crash

**To Reproduce**
1. Establish a peer connection
2. Received media stream with video track
3. Call the captureFrame function on the video track

**Expected behavior**
    • Frame got captured and return a ByteBuffer object

**Platform information**

* **Flutter version**: 
Flutter (Channel stable, 3.0.5, on macOS 12.6 21G115 darwin-x64, locale en-HK)
    • Flutter version 3.0.5 at /Users/alexlam/fvm/versions/3.0.5
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision f1875d570e (3 months ago), 2022-07-13 11:24:16 -0700
    • Engine revision e85ea0e79c
    • Dart version 2.17.6
    • DevTools version 2.12.2

* **Plugin version**: 0.9.8
* **OS**: Android
* **OS version**: Android 12, Android 13

**Error log**:
```
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x7ef013ea6a00
    rax 00007ef013ea6a00  rbx 0000000041e3e010  rcx 00000000001fa400  rdx 00000000001fa400
    r8  00000000001fa400  r9  00000000000023ab  r10 00000000f6d8fa17  r11 000000000000000d
    r12 0000000000000000  r13 00007ef01bcc29b0  r14 0000000000000000  r15 0000000000000001
    rdi 0000000041e3e010  rsi 00007ef013ea6a00
    rbp 00007ef013ea6a00  rsp 00007ef01a3e23d8  rip 00007ef01bcc29bc
backtrace:
      #00 pc 00000000004b39bc  /data/app/~~AMHDNxtV-53R55qFt7qDNQ==/${appId}-nqOh-SXVlWqBLOq7--Gitg==/base.apk!libjingle_peerconnection_so.so (BuildId: 629419e5b6984828)
      #01 pc 00000000004af7cb  /data/app/~~AMHDNxtV-53R55qFt7qDNQ==/${appId}-nqOh-SXVlWqBLOq7--Gitg==/base.apk!libjingle_peerconnection_so.so (BuildId: 629419e5b6984828)
[...]

PR #1227, opened 2023-01-11
Fix crash when calling captureFrame on a video track on android (#1129)

basyskom-jvoe:
The crash only happens on some Android devices (reproduced on Google Pixel 4).


cloudwebrtc:
lgtm

basyskom-jvoe:
Thanks for the merge!

PR #1243, opened 2023-01-20
Buffer overflow, while requesting device name

korvan:
When PlayoutDeviceName or RecordingDeviceName is used on linux, default device name can exceed default buffer size 128 Bytes. In this case WebRTC returns no null terminated buffer. And building string as
std::string(strNamUTF8) leads to usage of memory outside of buffer.
This unitialized memory can't then be decoded bu flutter code.


hiroshihorie (approved):
LGTM

PR #1246, opened 2023-01-23
Start audioswitch and only activate it when needed

mouEsam:
The problem:

No output devices are available unless a user media stream is created due to the switch manager not being present before the audio manager is started.
The solution:

Initialize and start the switch manager and only activate it when needed.


cloudwebrtc:
lgtm!

cloudwebrtc:
lgtm!

fullflash:
this is blocking device receiving audio from other apps.

mouEsam:
If by audio you mean the mic is not available to other apps, this doesn't appear to be happening. And if you mean no other application can play audio while the app is active, then that doesn't seem to be happening either.

fullflash:
sorry by bad  the problem is reducing music apps volume. i am starting spotify it plays with high quality then launching app and sound volume goes done and quality also. Only after i terminate the app the spotify sound goes to normal.

PR #1255, opened 2023-02-01
Added elinux support

Kofhein:
This commit adds elinux support, basically elinux is compatible with Windows API, so it doesn't require much, only CMake, add separate definition for _ELINUX and symlink as target platform is elinux-x64 and it would fail to find path ( linkage to libwebrtc.so)


Kofhein:
On main branch there's some issue which cause next exception when create offer called
[bad_optional_access.cc : 39] RAW: Bad optional access

On version 0.9.19 this is fine.

cloudwebrtc:
It may be that the header file does not match the version of libwebrtc.so.

BTW do you have cross-thread call exceptions thrown in debug mode?
https://github.com/flutter-webrtc/flutter-webrtc/pull/741#issue-1034311898

cloudwebrtc:
replace with  https://github.com/flutter-webrtc/flutter-webrtc/pull/1338

PR #1279, opened 2023-02-24
Fix: After the window is closed, the old source still exists

nikohpng:
When the sharing window is opened, the newly added page will be displayed in the window. However, after closing the application to be shared and reopening the sharing window, the application interface still exists


cloudwebrtc (approved):
lgtm!

Issue #1284, opened 2023-03-06
[iOS] Audio Output device switches to earpiece when mic is turner On and switch to speaker when mic turned off


rajansurani:
**Describe the bug**
When the user trie to turn ON the mic on **iOS devices**, the audio output switches from speaker to earpiece. And once the mic is turned OFF, audio output switches back to speaker.

Here is the similar issue reported on the live-kits flutter SDK.
https://github.com/livekit/client-sdk-flutter/issues/245

**To Reproduce**
Here is a test app, one can join from two iOS device and when the mic is turned on, the output device will be switched. 

https://testflight.apple.com/join/C1UOYbxh

**Expected behavior**
Audio output device should not switch when audio is toggled.

**Platform information**

* **Flutter version**:
 Flutter (Channel stable, 3.7.3, on macOS 13.1 22C65 darwin-arm64, locale
    en-IN)
    • Flutter version 3.7.3 on channel stable at
      /Users/dhiwise/Documents/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 9944297138 (4 weeks ago), 2023-02-08 15:46:04 -0800
    • Engine revision 248290d6d5
    • Dart version 2.19.2
    • DevTools version 2.20.1
* **Plugin version**:
* **OS**:  iOS
* **OS version**: 16.1.1

rajansurani:
Fixed in the version `0.0.24`

rajansurani:
@fullflash as per docs of AVAudioSession, when using video chat, it allows to use the output device suitable for video calls, and since earpiece is not one of that, it seems to resolve the issue.

shivanshtalwar0:
@rajansurani any update on #1285 ?

PR #1285, opened 2023-03-06
BugFix: iOS audio ouptut switching when toggling mic

rajansurani:
This is a PR to fix for Issue #1284.

Choosing a `AVAudioSessionModeVideoChat` as mode while using `PlayAndRecord` solves the issue of audio being switched to earpiece once the mic is toggled on and vice-versa.

### DIFF
--- a/common/darwin/Classes/AudioUtils.m
+++ b/common/darwin/Classes/AudioUtils.m
@@ -15,7 +15,7 @@
     config.category = AVAudioSessionCategoryPlayAndRecord;
     config.categoryOptions =
         AVAudioSessionCategoryOptionAllowBluetooth | AVAudioSessionCategoryOptionAllowBluetoothA2DP;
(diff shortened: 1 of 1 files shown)


cloudwebrtc (approved):
lgtm

shivanshtalwar:
With this change it doesn't give option to choose earpiece it defaults to speaker on ios call-kit call 👎🏻

shivanshtalwar0:
@rajansurani on the dart side how can we override it?

rajansurani:
> @rajansurani on the dart side how can we override it?

Still haven't got a perfect solution for both cases but probably shaning the mode of audio session when changing with Helper.speakerPhone could be an option.

fullflash:
looks like this implementation has been removed
in the current source there is no mode AVAudioSessionModeVideoChat

PR #1319, opened 2023-04-23
[Windows/Linux] `RTPtransceiver::getCurrentDirection` returns correct value

amantoux:
libwebrtc implementation wasn't calling `RTCRtpTransceiver::current_direction`


cloudwebrtc (approved):
lgtm!

PR #1329, opened 2023-04-30
fix audio mode

igala:
This will allow to route to earpiece and speaker.


UserJah:
Is this anti fix of https://github.com/flutter-webrtc/flutter-webrtc/pull/1285 ?

Issue #1345, opened 2023-05-10
Screen Sharing - Using Broadcast Extension to share any app doesn't work


ipainchaud:
There are two different ways to share the screen on iOS described [here](https://github.com/flutter-webrtc/flutter-webrtc/wiki/iOS-Screen-Sharing): 

1. In-App
2. Broadcast Extension

I followed all the instructions described [here](https://github.com/flutter-webrtc/flutter-webrtc/wiki/iOS-Screen-Sharing) without success. It doesn't crash, doesn't give any error message, just doesn't seem to return the MediaStream containing the video track to transmit.

When I use the In-App screen sharing, no issue.

**To Reproduce**
Follow steps described [here](https://github.com/flutter-webrtc/flutter-webrtc/wiki/iOS-Screen-Sharing).

**Expected behavior**
The Screen should be shared.

* **Flutter version**: 
$ flutter doctor -v
[✓] Flutter (Channel stable, 3.7.12, on macOS 13.3.1 22E772610a darwin-x64, locale en-US)
    • Flutter version 3.7.12 on channel stable at /usr/local/Caskroom/flutter/3.0.2/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 4d9e56e694 (3 weeks ago), 2023-04-17 21:47:46 -0400
    • Engine revision 1a65d409c7
    • Dart version 2.19.6
    • DevTools version 2.20.1

[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.0)
    • Android SDK at /Users/israelpainchaud/Library/Android/sdk
    • Platform android-33, build-tools 33.0.0
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-17.0.6b802.4-9586694)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 14.3)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 14E222b
    • CocoaPods version 1.12.1

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2022.2)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
[...]

PR #1347, opened 2023-05-10
Fix: broadcast extension not found fallback logic

ipainchaud:
fixing #1345.

It turns out that the issue is caused by a bad "fallback" if the `RTCScreenSharingExtension` is not found. The instructions found [here](https://github.com/flutter-webrtc/flutter-webrtc/wiki/iOS-Screen-Sharing) don't mention that we should set the extension id there.

In all cases though, it would be nice to update the instructions found [here](https://github.com/flutter-webrtc/flutter-webrtc/wiki/iOS-Screen-Sharing).

With this solution, if the key is not found, the broadcast selector lists what I suspect to be all the other available extensions.


cloudwebrtc (approved):
lgtm

PR #1385, opened 2023-06-27
Fix setMicrophoneMute() not awaitable

aschulz90:
`setMicrophoneMute()` is async but didn't return a Future, which makes it not usable with `await`.


cloudwebrtc (approved):
lgtm!

PR #1393, opened 2023-07-07
Update build.gradle

kBrahman:
Fix Gradle 8 compatibility issue


kBrahman:
Fixed Gradle 8 compatibility issue

Issue #1241, opened 2023-01-20
[Linux] libwebrtc.so: no such file or directory on destination PCs


nikitatg:
**Description**
On destination PC with no flutter environment installed I'm unpacking a "Release" bundle to ~/apps/MyApp/ grant rights to run executable and when trying to run app, I'm getting the following error and app crash.

`/home/user1/apps/MyApp/my_app_executable: error while loading shared libraries: libwebrtc.so: cannot open shared object file: no such file or directory`

Meanwhile there is no any problems on developer PC, where app was built

Note that libwebrtc.so lies peacefully near the executable at /home/user1/apps/MyApp/lib with other plugin's libs

**To Reproduce**

- Create dummy flutter project
- Add pubspec dependency flutter_webrtc: 0.9.16 then run `flutter pub get`
- Build app with `flutter build linux`
- Copy a "Release" bundle to any Linux PC with no flutter environment installed
- Grant execute privilege on bundle executable
- Try to run executable from terminal

**Platform information**

**Flutter version**: 
<details>
  <summary>flutter doctor -v</summary>
[√] Flutter (Channel stable, 3.3.10, on Microsoft Windows [Version 10.0.19044.2486], locale ru-RU)
    • Flutter version 3.3.10 on channel stable at C:\Work\flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 135454af32 (5 weeks ago), 2022-12-15 07:36:55 -0800
    • Engine revision 3316dd8728
    • Dart version 2.18.6
    • DevTools version 2.15.0

[X] Android toolchain - develop for Android devices
    • Android SDK at C:\Users\agnik\AppData\Local\Android\sdk
    X cmdline-tools component is missing
      Run `path/to/sdkmanager --install "cmdline-tools;latest"`
      See https://developer.android.com/studio/command-line for more details.

[√] Chrome - develop for the web
    • Chrome at C:\Program Files\Google\Chrome\Application\chrome.exe

[√] Visual Studio - develop for Windows (Visual Studio Community 2019 16.11.19)
    • Visual Studio at C:\Program Files (x86)\Microsoft Visual Studio\2019\Community
    • Visual Studio Community 2019 version 16.11.32901.82
    • Windows 10 SDK version 10.0.19041.0

[!] Android Studio (version 2022.1)
    • Android Studio at C:\Program Files\Android\Android Studio
    • Flutter plugin can be installed from:
[...]

PR #1398, opened 2023-07-14
fix: libwebrtc.so: no such file or directory 

KoiFresh:
Adds $ORIGIN to the RPATH of the plugin, to be able to find the libwebrtc.so file after distributing the bundle. 

will solve: 
#1212
#1241


cloudwebrtc (approved):
lgtm!

PR #1401, opened 2023-07-23
feat: expose android audio modes

davidliu:
default audio modes is also changed from MODE_IN_COMMUNICATION to MODE_NORMAL.


cloudwebrtc (approved):
lgtm

PR #1406, opened 2023-08-01
Partially rebuild RTCVideoView when renderVideo value changes

ycherniavskyi:
When the `srcObject` value changes (and as a consequence, the `renderVideo` value of `videoRenderer` changes) in a place not related to the widget without a `setState` call (for example within a bloc), the associated `RTCVideoView` widget does not rebuild. As a result, it continues to display a placeholder or Container widget, even if `RTCVideoRenderer.renderVideo` is `true`.

In this pull request, I expand the usage of the `ValueListenableBuilder` widget, which already exists within `RTCVideoView` widget. This allows the actual `renderVideo` value to be respected and triggers a partial rebuild of the `RTCVideoView` widget when the value changes.

By the way in `RTCVideoView` for web this done by [`_onRendererListener`](https://github.com/flutter-webrtc/flutter-webrtc/blob/fad3b57203096516183752e14f4901e893699984/lib/src/web/rtc_video_view_impl.dart#L46) callback that call `setState`.

### DIFF
--- a/lib/src/native/rtc_video_view_impl.dart
+++ b/lib/src/native/rtc_video_view_impl.dart
@@ -49,19 +49,19 @@ class RTCVideoView extends StatelessWidget {
                 return SizedBox(
                   width: constraints.maxHeight * value.aspectRatio,
                   height: constraints.maxHeight,
-                  child: child,
+                  child: Transform(
+                    transform: Matrix4.identity()..rotateY(mirror ? -pi : 0.0),
+                    alignment: FractionalOffset.center,
+                    child: value.renderVideo
+                        ? Texture(
+                            textureId: videoRenderer.textureId!,
+                            filterQuality: filterQuality,
+                          )
+                        : child,
+                  ),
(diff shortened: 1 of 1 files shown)


wer-mathurin (approved):
LGTM

cloudwebrtc (approved):
lgtm

cloudwebrtc:
This PR seems to cause flutter Windows (win 10) to not work correctly. To reproduce the steps, run getUserMedia or LoopBack Sample in the example, the randomness of the video rendering component cannot be displayed, re-vs2022 debugging, flutter_video_renderer.cc::onFrame is called, but CopyPixelBuffer is not called by flutter, It seems that Texture is not rendering correctly in flutter UI.

cloudwebrtc:
I'm out of ideas, I tried some modifications, and only by reverting these changes and the Windows app can work, so I need to temporarily revert this PR until the issue with Windows rendering is resolved.

PR #1408, opened 2023-08-06
fix: fix android earpiece not being replaced after wired headset is disconnected

davidliu:
Original issue: https://github.com/livekit/client-sdk-android/issues/251


cloudwebrtc (approved):
lgtm

Issue #351, opened 2020-09-18
Hangup doesn't close the audio stream


LucMoreau33560:
Hello,

**Describe the bug**
When I hangup, the streams are not closed. I can continue to dialog between my two devices

**To Reproduce**
The simplest way is to install flutter webrtc demo and select P2P Call Sample.
The signaling server is flutter-webrtc-server.
my hangup function : 
_hangUp() {
    if (_signaling != null) {
      _signaling.bye();
      _localRenderer.srcObject = null;
      _remoteRenderer.srcObject = null;
      _inCalling = false;
    }
  }
And in the signaling onstate change :
  case SignalingState.CallStateBye:
            this.setState(() {
              _localRenderer.srcObject = null;
              _remoteRenderer.srcObject = null;
              _inCalling = false;
  });
            break;

The signaling output when I hangup : 

2020-09-18T08:14:03+02:00 INF Send keepalive !!!
2020-09-18T08:14:03+02:00 INF Send data: {}
2020-09-18T08:14:06+02:00 INF Recivied data: {"type":"bye","data":{"session_id":"L56789-C56789","from":"C56789"}}
2020-09-18T08:14:06+02:00 INF Send data: {"data":{"session_id":"L56789-C56789","to":"L56789"},"type":"bye"}
2020-09-18T08:14:06+02:00 INF Send data: {"data":{"session_id":"L56789-C56789","to":"C56789"},"type":"bye"}
2020-09-18T08:14:08+02:00 INF Send keepalive !!!
2020-09-18T08:14:08+02:00 INF Send data: {}

**Expected behavior**
When I hangup, I expect the stream to be closed.

**Platform information**

* **Flutter version**: 
flutter doctor -v
[✓] Flutter (Channel stable, 1.20.3, on Mac OS X 10.15.6 19G73, locale fr-FR)
    • Flutter version 1.20.3 at /Users/luc/dev/flutter
    • Framework revision 216dee60c0 (il y a 2 semaines), 2020-09-01 12:24:47
      -0700
    • Engine revision d1bc06f032
    • Dart version 2.9.2

 
[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.0)
    • Android SDK at /Users/luc/Library/Android/sdk
    • Platform android-30, build-tools 30.0.0
    • Java binary at: /Applications/Android
      Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build
[...]

PR #1410, opened 2023-08-08
Fix: [Android] Expose helper to clearCommunicationDevice on AudioManager.AUDIOFOCUS_LOSS

chrisflutterdev:
- Expose helper method to clearCommunicationDevice
- Add case: AndroidAudioFocusMode.loss 

Android documentation recommends that when an app finishes a call or session, we should clear the active communication device: https://developer.android.com/guide/topics/connectivity/ble-audio/audio-manager#end_the_session
 
In addition, we should tell the Audio Manager that we want to gain and lose the audio focus:
https://developer.android.com/reference/kotlin/androidx/media/AudioManagerCompat#AUDIOFOCUS_GAIN()

This fix will help to ensure that the user can return to their original audio stream when a webrtc session is finished. 
May resolve https://github.com/flutter-webrtc/flutter-webrtc/issues/351


cloudwebrtc (approved):
lgtm

PR #1415, opened 2023-08-11
feat: support force SW codec list for android, and disable HW codec for VP9 by default

cloudwebrtc:


android/src/main/java/com/cloudwebrtc/webrtc/MethodCallHandlerImpl.java
```diff
@@ -170,18 +174,21 @@ private void initialize(int networkIgnoreMask, boolean forceSWCodec) {
...
 
-    if (forceSWCodec) {
-      factoryBuilder
-              .setVideoEncoderFactory(new SoftwareVideoEncoderFactory())
-              .setVideoDecoderFactory(new SoftwareVideoDecoderFactory());
-    } else {
-      // Initialize EGL contexts required for HW acceleration.
-      EglBase.Context eglContext = EglUtils.getRootEglBaseContext();
+    // Initialize EGL contexts required for HW acceleration.
+    EglBase.Context eglContext = EglUtils.getRootEglBaseContext();
```
MohamedRisaldarTA, line 178: can we avoid initialising EGL context when SW codec is used.
It reduces the ability to use more streams.

cloudwebrtc: Yes, I think it's possible, would you like to create a PR for it? I will review it soon

MohamedRisaldarTA: I am little busy on other things now,
Let me see if I can get some free time by the end of september if it's not done earlier.

Issue #811, opened 2021-12-16
bluetooth audio routing issue 


denkoul:
**Describe the bug**
With bluetooth headphones connected on **Android** phone, the initial audio output is set to bluetooth. When I change it to the loudspeaker and again back to bluetooth using:

```
      if (peerConnection != null) {
        for (MediaStream? mediaStream in peerConnection!.getRemoteStreams()) {
          if (mediaStream != null)
            mediaStream.getAudioTracks().forEach(
              (element) {
                element.enableSpeakerphone(enabled);
              },
            );
        }
      } 
```

the sound is not routing back to bluetooth, but it plays through earpiece speaker instead.

**To Reproduce**

- Start call with connected bluetooth
- Change to loudspeaker
- Switch back to bluetooth
- No sound through bluetooth 

**Expected behavior**

- Start call with connected bluetooth
- Change to loudspeaker
- Switch back to bluetooth
- Sound should play in bluetooth

**Platform information**
Android

* **Flutter version**: 

```
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 2.8.0, on macOS 12.0.1 21A559 darwin-x64, locale en-GR)
[✓] Android toolchain - develop for Android devices (Android SDK version 32.0.0-rc1)
[✓] Xcode - develop for iOS and macOS (Xcode 13.2)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2020.3)
```

* **Plugin version**: 0.8.0
* **OS**:  Android 
* **OS version**: 9 and 10

pannawatth:
Hi @denkoul, Have you found any solutions?

PR #1417, opened 2023-08-11
Fix/issue for audio device switch

cloudwebrtc:


Andrushen:
Hey! Does it fix https://github.com/flutter-webrtc/flutter-webrtc/issues/811?
Pretty old bug, but still no updates on it

PR #1460, opened 2023-09-27
fix "unlock of unowned mutex" error when call "captureFrame()" func on windows

liliBestCoder:
When i use flutter-webrtc plugin build my app on window 10, i got this error like belowing

"D:\a_work\1\s\src\vctools\crt\github\stl\src\mutex.cpp(164): unlock of unowned mutex".

so i had to read some relative source code, i found some problems in flutter_frame_capturer.cc file, i fix them and it work


cloudwebrtc (approved):
lgtm

PR #1475, opened 2023-10-24
Fix bluetooth sco not stopping after all peer connections disconnect

davidliu:
Updates AudioSwitch lib to fix this issue.


cloudwebrtc (approved):
lgtm

PR #1479, opened 2023-10-31
fix: revert changes for fix simulcast bug on android.

cloudwebrtc:


davidzhao:
what should be the correct ordering of sendEncodings? I think it should not flip the ordering right? i.e. whatever the user is setting, it should maintain that order

cloudwebrtc:
This may be an ineffective fix, I reverted this modification, and subscribers can still switch to 720p correctly when the network quality is good, so it can only be said that in some cases the network is really bad, and switching the video quality manually becomes invalid. I'm already confused xD

PR #1508, opened 2023-12-27
fix android wrong screen capture size.

Jaemin-VIRNECT:
When you use the android screen capture feature, it provides an incorrect capture size; in most cases, this bug may not be a significant issue. However, it becomes problematic when working with the Flutter overlay view. (iOS works okay.)

The log is shown below.
2023-12-27 14:56:27.549 14766-14766 FlutterWebRTCPlugin D OrientationAwareScreenCapturer.startCapture: 1080x2122@30
Test Device: Galaxy S20 Ultra 5G
Device screen size: 1080x2316


cloudwebrtc (approved):
lgtm

PR #1514, opened 2024-01-24
Supports ensureAudioSsession method for iOS only.

Jaemin-VIRNECT:
Sometimes this needs to be called manually when changing the audioSession configuration with a 3rd party plugin.


cloudwebrtc (approved):
lgtm

PR #1552, opened 2024-03-19
support media projection feature for android 14

Jaemin-VIRNECT:
After AOS 14, changeCaptureFormat() should resize the VirtualDisplay instead of recreating it due to the limitation on reusing MediaProjectionPermissionResultData. And it is necessary to obtain capture permission first and request foreground service; otherwise, the app crashes due to the media projection permission policy.
Use Helper.requestCapturePermission() to obtain capture permission on Android 14 specially targetSdkVersion 34 or above.

Tested on Android 13, 14 and targetSdkVersion 33, 34 combinations.


cloudwebrtc (approved):
lgtm

PR #1568, opened 2024-04-23
fix(iOS): remember speakerphone mode for ensureAudioSession

holzgeist:
otherwise event like "onTrackAdd" will call setSpeakerphoneOn even though I called setSpeakerphoneOnButPeferBluetooth before.


holzgeist:
@cloudwebrtc I can't assign you as reviewer, so here's a ping that this PR is ready for review

cloudwebrtc (approved):
lgtm!

PR #1591, opened 2024-05-30
Feat/libwebrtc m125

cloudwebrtc:


rantoniozzi:
Hi with this revision using Helper.setMicrophoneMute the first time work and the audio is not available,  but when i try to reactivate by setting Helper.setMicrophoneMute to false the audio is not available.
For me it work until the version 0.10.8, which is the last that don't use libwebrtc to m125

PR #1603, opened 2024-06-17
Supports video capturer stop and restart features on Android.

Jaemin-VIRNECT:
Supports video capturer stop and restart features on Android. You must manually stop and restart the video capturer. Otherwise, after accessing the camera in another app, the camera won't reopen when you return to the app.


Jaemin-VIRNECT:
@cloudwebrtc 
Would you please consider merging it into the main branch?. Thanks.

victortive:
Shouldn't the app call dispose on the renderers instead?

PR #1604, opened 2024-06-17
Fix AOS 14 virtual display size bug.

Jaemin-VIRNECT:
When changing the device's orientation, the virtual display size does not resize to the given value, causing the sent video to appear somewhat distorted.


cloudwebrtc (approved):
lgtm!

PR #1612, opened 2024-06-28
Fixes #1611, Resolve running into a NullPointerException on finding RtpSender for sendDTMF

### LINKED ISSUE #1611: NullPointerException on sendDTMF: Attempt to invoke virtual method 'java.lang.String org.webrtc.MediaStreamTrack.kind()' on a null object reference
**Describe the bug**
The onMethodCall case "sendDTMF" in MethodCallHandlerImpl.java can result in a NullPointerException preventing DTMF tones to be sent.

`PlatformException(error, Attempt to invoke virtual method 'java.lang.String org.webrtc.MediaStreamTrack.kind()' on a null object reference, null, java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String org.webrtc.MediaStreamTrack.kind()' on a null object reference`

**To Reproduce**
Using the flutter-webrtc/dart-sip-ua example app, receive an incoming call and try to send DTMF.
This is the only way I consistently end up with a NULL in the RtpSender list from PeerConnection.

**Expected behavior**
The method finds a RtpSender and its MediaStreamTrack that are not null and of kind "audio" to use the attached DtmfSender as the DTMF Sender.

**Platform information**

* **Flutter version**: 3.22.0
* **Plugin version**: 0.11.1
* **OS**: Android
* **OS version**: Android 14

lineaira:
Check if RtpSender and its MediaStreamTrack from PeerConnection are not null when looking for a MediaStreamTrack of kind audio when finding a RtpSender to call insertDtmf in onMethodCall case: 'sendDTMF'.

### DIFF
--- a/android/src/main/java/com/cloudwebrtc/webrtc/MethodCallHandlerImpl.java
+++ b/android/src/main/java/com/cloudwebrtc/webrtc/MethodCallHandlerImpl.java
@@ -386,7 +386,7 @@ public class MethodCallHandlerImpl implements MethodCallHandler, StateProvider {
           RtpSender audioSender = null;
           for (RtpSender sender : peerConnection.getSenders()) {
 
-            if (sender.track().kind().equals("audio")) {
+            if (sender != null && sender.track() != null && sender.track().kind().equals("audio")) {
               audioSender = sender;
             }
           }


cloudwebrtc (approved):
lg!

PR #1636, opened 2024-07-22
Report actual sizes for camera media stream track and remove adaptOutputFormat usage

davidliu:
MediaStreamTrack settings should report the actual values rather than the target values given by constraints. 

Also remove adaptOutputFormat due it causing extreme downscaling.

Addresses https://github.com/webrtc-sdk/webrtc/issues/133


cloudwebrtc (approved):
lg!

PR #1670, opened 2024-09-20
render cameras into web canvas

jezell:


jezell:
Updated to further optimize memory usage, and work around a bug in the flutter 3.24 renderer that causes Flutter CanvasKit to run out of memory when using HtmlElementView or platform layers.

dganzella:
damn, thats good stuff

I think I can easily add bodypix segmentation on top of the canvas, right?

Just customizing the _ImageFlipPainter  code.

I also think I'm gonna implement enabling or disabling the usage of canvas per-video, as to improve performance when not using any effects like blur

dganzella:
Oh. This is just local. The canvas streaam is not being sent

theniceboy:
Hey @jezell , I was unable to get your PR to work. My `RTCVideoView` would just render a black screen. This is a minimal example:

```dart
class SelfieCamera extends StatefulWidget {
  @override
  _SelfieCameraState createState() => _SelfieCameraState();
}

class _SelfieCameraState extends State<SelfieCamera> {
  final _localRenderer = RTCVideoRenderer();
  MediaStream? _localStream;
  bool _initialized = false;

  @override
  void initState() {
    super.initState();
    _initializeCamera();
  }

  Future<void> _initializeCamera() async {
    await _localRenderer.initialize();

    // Get user media with video only
    final Map<String, dynamic> mediaConstraints = {
      'audio': false,
      'video': {
        'facingMode': 'user',
      }
    };

    try {
      _localStream = await navigator.mediaDevices.getUserMedia(mediaConstraints);
      _localRenderer.srcObject = _localStream;
      setState(() {
        _initialized = true;
      });
    } catch (e) {
      print('Error accessing camera: $e');
    }
  }

  @override
  void dispose() {
    _localRenderer.dispose();
    _localStream?.getTracks().forEach((track) => track.stop());
    super.dispose();
  }

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      body: _initialized
          ? RTCVideoView(
              _localRenderer,
              mirror: true,
              objectFit: RTCVideoViewObjectFit.RTCVideoViewObjectFitCover,
            )
          : const Center(
              child: CircularProgressIndicator(),
            ),
    );
  }
}
```

```yaml
dependency_overrides:
  flutter_webrtc:
    git:
      url: 'https://github.com/jezell/flutter-webrtc'
      ref: 'e34a9fb73c6b01d799ee79cc9ab435cf3be06bac'
  dart_webrtc:
    git:
      url: 'https://github.com/jezell/dart-webrtc'
      ref: 'a6a18675e48fc7d6b89aa798a18283add17a6b60'
  livekit_client:
    git:
      url: 'https://github.com/jezell/client-sdk-flutter'
      ref: '1f542c473de75b6d0dae8da094fa7eb74f12a882'
```

Please let me know if you can reproduce it. Thanks!

jezell:
Added your sample code to a project here:

https://github.com/jezell/test-flutter-webrtc

Has all the dependency overrides and seems to work for me. Do you get the same result with that project?

jezell:
@dganzella Bodypix segmentation would be a really cool addition! I don't see any reason it couldn't be done. We did something similar with livekit in js (I think there is even a js plugin for livekit that does this, but it's not ported to flutter yet).

jezell:
@dganzella this is a good reference for background blur if that's what you are going for:

https://github.com/Volcomix/virtual-background

theniceboy:
> Added your sample code to a project here:
> 
> https://github.com/jezell/test-flutter-webrtc
> 
> Has all the dependency overrides and seems to work for me. Do you get the same result with that project?

Thank you for providing the example. Your example worked for me, but when I compared it to mine, I found out the reason why it didn't work for me. I had
```html
<style>
	.container {
		height: 100vh;
	}
</style>
```
in my `<head>` and
```html
<div id="loading_indicator" class="container">
</div>
```
in the `<body>` of my `index.html`. For whatever reason, it causes the `RTCVideoView` to render just black. What in the world is going on?😂

jezell:
The browser is a little finicky about video element rendering. If it thinks the video element itself that the texture is being pulled from isn't visible, it will not decode the frames. That's probably what you were seeing.

theniceboy:
I see. That makes sense. After some testing, it seems like the video views were rendering in some parts of my app but not others. It might be good to note in this PR that it could pose some limitations on where you can place this video view.

jezell:
If you can throw up a sample on github that illustrates the issue I'll take a look and put together an update

theniceboy:
Any updates on this?

PR #1692, opened 2024-10-29
fix: race condition during video renderer initialization

holzgeist:
While dart is single threaded, async code can still run concurrently. Execution can be yielded at any usage of `await`. This means that, when calling renderer.initialize twice in relatively short amount of time, it will allocate two textures, but only free one of them.

This commit fixes this by atomically (no `await` in between) checking and setting an initialization completer.
The completer will wait for pending initializations or (more likely) return immediately if already initialized


cloudwebrtc (approved):
lg！

PR #1695, opened 2024-10-31
feat: set facing mode after selecting camera

holzgeist:
This allows accessing facing mode from track settings even if camera was selected by ID or other constraints

### DIFF
--- a/android/src/main/java/com/cloudwebrtc/webrtc/GetUserMediaImpl.java
+++ b/android/src/main/java/com/cloudwebrtc/webrtc/GetUserMediaImpl.java
@@ -751,6 +751,13 @@ class GetUserMediaImpl {
         deviceId = result.first;
         VideoCapturer videoCapturer = result.second;
 
+        if (facingMode == null && cameraEnumerator.isFrontFacing(deviceId)) {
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/cloudwebrtc/webrtc/GetUserMediaImpl.java
```diff
@@ -751,6 +751,13 @@ private ConstraintsMap getUserVideo(ConstraintsMap constraints, MediaStream medi
         deviceId = result.first;
         VideoCapturer videoCapturer = result.second;
 
+        if (facingMode == null && cameraEnumerator.isFrontFacing(deviceId)) {
```
holzgeist, line 754: I'm not sure if we should check `facingMode == null` here and below. Is it possible to set a facingMode constraint and get a camera that doesn't fulfil it?

holzgeist:
This should be similar as the darwin implementation: https://github.com/flutter-webrtc/flutter-webrtc/blob/main/common/darwin/Classes/FlutterRTCMediaStream.m#L487-L491
But I'm not too familiar with ObjC

PR #1726, opened 2024-12-19
feat: Migrate to the new Surface API.

cloudwebrtc:


zhuhaichao518:
@cloudwebrtc I forked very old version and updated the whole android folder recently. I am not very familiar with android texture register, but I see the video not rendered after resume from background. I guess we need to set `surface = null;` at the end of surfaceDestroyed() function? That fixes the issue in my case.

PR #1764, opened 2025-02-12
feat: Add option to start capture without broadcast picker

ladvoc:
When using the broadcast capturer on iOS, the `deviceId` constraint can now be set to `broadcast-manual` instead of `broadcast` to allow capture to begin without presenting the broadcast picker. When using this option, the user is responsible for managing the broadcast state externally. 

This change is non-breaking.


cloudwebrtc (approved):
lg!

PR #1801, opened 2025-03-27
fix: is not abstract and does not override abstract method onSurfaceCreated() in Callback

gperez-88:
fix: is not abstract and does not override abstract method onSurfaceCreated() in Callback


cloudwebrtc:
will use thid PR instead.
https://github.com/flutter-webrtc/flutter-webrtc/pull/1801

PR #1808, opened 2025-04-03
[Android] fix: remove setPreferredInputDevice when getUserAduio.

cloudwebrtc:
Calling `setPreferredInputDevice` will cause HW AEC/NS failure on some devices, causing echo during voice calls. Known devices include Galaxy S25 Ultra, Galaxy Z Fold6.

In addition, we should avoid using [Helper.selectAudioInput](https://github.com/flutter-webrtc/flutter-webrtc/blob/main/lib/src/helper.dart#L108) during a call to avoid causing echo cancellation failure on these devices.


ahmetakil:
Can you share more about what alternatives we can use instead of Helper.selectAudioInput, for instance to switch to a wireless device, and then switch back to the devices microphone itself

Issue #1744, opened 2025-01-15
Failed to handle method call - DeviceOrientationManager


GuillianDrouin:
**Describe the bug**

I have a problem only on Android, it appeared between version 0.12.3 which works and 0.12.4 which is KO for me.
I tested on Samsung fold and Pixel 6.

```
E/MethodChannel#FlutterWebRTC.Method(23426): Failed to handle method call
E/MethodChannel#FlutterWebRTC.Method(23426): java.lang.NullPointerException: Attempt to invoke virtual method 'android.content.Intent android.app.Activity.registerReceiver(android.content.BroadcastReceiver, android.content.IntentFilter)' on a null object reference
E/MethodChannel#FlutterWebRTC.Method(23426): 	at com.cloudwebrtc.webrtc.video.camera.DeviceOrientationManager.start(DeviceOrientationManager.java:57)
E/MethodChannel#FlutterWebRTC.Method(23426): 	at com.cloudwebrtc.webrtc.video.camera.CameraUtils.<init>(CameraUtils.java:48)
E/MethodChannel#FlutterWebRTC.Method(23426): 	at com.cloudwebrtc.webrtc.MethodCallHandlerImpl.initialize(MethodCallHandlerImpl.java:175)
E/MethodChannel#FlutterWebRTC.Method(23426): 	at com.cloudwebrtc.webrtc.MethodCallHandlerImpl.onMethodCall(MethodCallHandlerImpl.java:343)
E/MethodChannel#FlutterWebRTC.Method(23426): 	at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler.onMessage(MethodChannel.java:267)
E/MethodChannel#FlutterWebRTC.Method(23426): 	at io.flutter.embedding.engine.dart.DartMessenger.invokeHandler(DartMessenger.java:292)
E/MethodChannel#FlutterWebRTC.Method(23426): 	at io.flutter.embedding.engine.dart.DartMessenger.lambda$dispatchMessageToQueue$0$io-flutter-embedding-engine-dart-DartMessenger(DartMessenger.java:319)
E/MethodChannel#FlutterWebRTC.Method(23426): 	at io.flutter.embedding.engine.dart.DartMessenger$$ExternalSyntheticLambda0.run(Unknown Source:12)
E/MethodChannel#FlutterWebRTC.Method(23426): 	at android.os.Handler.handleCallback(Handler.java:942)
[...]

Issue #1799, opened 2025-03-25
Unable to getUserMedia: Attempt to invoke virtual method 'android.content.Intent android.app.Activity.registerReceiver(android.content.BroadcastReceiver, android.content.IntentFilter)' on a null object reference


devKaran-coder:
**I am using flutter_webrtc plugin from sip_ua with flutter_foreground_task in between when I call this function which is mention below I got this error**

* **Flutter version**: 3.29.0
* **Plugin version**: flutter_webrtc-0.12.12+hotfix.1
* **OS**: Android

_funcForCalling(String number) async {
  try {
    final mediaConstraints = <String, dynamic>{'audio': true, 'video': false};

    webrtc.MediaStream mediaStream;

    mediaConstraints['video'] = false;
    mediaStream =
        await webrtc.navigator.mediaDevices.getUserMedia(mediaConstraints);

    helper!.call(number, mediaStream: mediaStream);
  } catch (e, s) {
    print("error is $e");
    print("stack stace is $s");
  }
}
 E/MethodChannel#FlutterWebRTC.Method(31010): Failed to handle method call
E/MethodChannel#FlutterWebRTC.Method(31010): java.lang.NullPointerException: Attempt to invoke virtual method 'android.content.Intent android.app.Activity.registerReceiver(android.content.BroadcastReceiver, android.content.IntentFilter)' on a null object reference
E/MethodChannel#FlutterWebRTC.Method(31010): 	at com.cloudwebrtc.webrtc.video.camera.DeviceOrientationManager.start(DeviceOrientationManager.java:57)
E/MethodChannel#FlutterWebRTC.Method(31010): 	at com.cloudwebrtc.webrtc.video.camera.CameraUtils.<init>(CameraUtils.java:48)
E/MethodChannel#FlutterWebRTC.Method(31010): 	at com.cloudwebrtc.webrtc.MethodCallHandlerImpl.initialize(MethodCallHandlerImpl.java:176)
E/MethodChannel#FlutterWebRTC.Method(31010): 	at com.cloudwebrtc.webrtc.MethodCallHandlerImpl.onMethodCall(MethodCallHandlerImpl.java:345)
E/MethodChannel#FlutterWebRTC.Method(31010): 	at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler.onMessage(MethodChannel.java:267)
[...]

PR #1840, opened 2025-05-19
fix: calls in terminated mode by disabling orientation manager

td-famedly:
I don't have a better solution at the moment, but `handleOrientationChange` looks empty so this should be ok.

Fixes: https://github.com/flutter-webrtc/flutter-webrtc/issues/1744, https://github.com/flutter-webrtc/flutter-webrtc/issues/1799


cloudwebrtc (approved):
lg

PR #1879, opened 2025-07-08
Ensure both video and audio tracks are added before starting the muxer

jawad1257:
Previously, the code started the muxer as soon as either the audio or video track was added, which could cause the app to crash. The updated logic ensures that the muxer only starts after both the audio and video tracks have been added.


cloudwebrtc (approved):

jawad1257:
@cloudwebrtc  can you merge this?

Issue #1677, opened 2024-10-04
Flutter webRtc  record audio only not working


vijaymsc:
```
  Future<void> startRecording() async {
    isRecording = !isRecording;
    setState(() {});
    if (isRecording) {
      try {
        _mediaRecorder = await MediaRecorder();
         await _mediaRecorder!.start(
          '/sdcard/temp/records/${DateTime.now()}-${widget.calleeId}.wav',
           audioChannel: RecorderAudioChannel.INPUT,
           // videoTrack: _mediaStream.getVideoTracks().first,
         );
     
      } catch (e, s) {
         _mediaRecorder!.stop();
        _mediaRecorder = null;
        if(mounted){
          ScaffoldMessenger.of(context).showSnackBar( SnackBar(content: Text(e.toString())));
        }
        print('Error while start recording');
      }
    } else {
      if (_mediaRecorder == null) return;
      try {
        await _mediaRecorder?.stop();
        _mediaRecorder = null;
      } catch (e) {
        print('Error while start stop  recording');
      }
    }
  }
```

error Log

Video track is null D/FlutterWebRTCPlugin(26651): 
startRecordToFile(): Audio-only recording not implemented yet

Issue #1501, opened 2023-12-13
Audio-only recording not implemented yet


patelnirav48:
Getting error "Audio-only recording not implemented yet"
Any idea when this feature is available?

@rostopira

rostopira:
Unfortunetaly, I don't have time for open source right now

patelnirav48:
@rostopira Can you give some reference where we can implement by own? Would be great help. Thanks

rostopira:
@patelnirav48 you can take a look at my old attempts to do that: https://github.com/flutter-webrtc/flutter-webrtc/pull/1189

patelnirav48:
@rostopira Thanks for your help. However i want to give support of audio only recoding (Here https://prnt.sc/Ux05adLm-Srv)
In above I couldn't able to find regards to this for android. Do you give some idea how we can achieve audio recording only?

vijaymsc:
Did it got fixed @patelnirav48  sir? I am too facing the same issue

niravp-cliqbux:
> Did it got fixed @patelnirav48 sir? I am too facing the same issue

No

alpha2048:
Same here. Any updates?

PR #1884, opened 2025-07-17
Add audio recording for Android Platform

maloleroy:
Hi, I've been working with @Aul16 on a project that requires audio-only recording using the [sip_ua](https://github.com/flutter-webrtc/dart-sip-ua) package. He implemented it for the Android platform in Java and after thorough testing, it works as a charm. It produces a `MP3` file.

Here is some demo code that is now possible on a SIP/WebRTC audio call:
```dart
// On call start
final recorder = MediaRecorder();
await recorder.start(
  path,
  audioChannel: RecorderAudioChannel.INPUT, // or RecorderAudioChannel.OUTPUT
);
// On call stop
await recorder.stop();
```

Here, "on call start" means in `sip_ua` something like
```dart
call.peerConnection!.onAddStream = (MediaStream stream) async {
  // Add a small delay to ensure the stream is fully established
  await Future.delayed(const Duration(milliseconds: 500));
  await startRecording(); // apparently the stream parameter is not needed
}
```
and "on call stop" means something like
```dart
// in a SipUaHelperListener's callStateChanged
if (state.state == CallStateEnum.FAILED || state.state == CallStateEnum.ENDED) {
  stopRecording();
}
```

We also tested using multiple `MediaRecorder` instances at once, one on `RecorderAudioChannel.INPUT` and one on `RecorderAudioChannel.OUTPUT` (for the both sides of a phone call), and it apparently doesn't cause any problems.

Note that we didn't make any automated tests and that there is still no implementation for other platforms than Android.


maloleroy:
While this PR doesn't exactly solve these issues, I'm mentioning #1677 and #1501 here

cloudwebrtc (approved):
lgtm!

PR #1900, opened 2025-07-25
RECORDINGS - Add fallback resolutions for unsupported stream frame sizes on low-end Android devices

daniel-g-favoreto-opl:
This PR adds fallback resolutions to handle cases where the default stream frame resolution is not supported by the device. This issue typically occurs on low-end Android devices that lack support for higher resolutions. With this change, the system will automatically attempt to use compatible resolutions to ensure stream encoding continues without crashing.

Example:
If the original stream resolution is QHD and the encoder fails to configure it, the system will automatically fallback to FHD and try again. This process continues down the list of fallback resolutions until one is successfully configured.


bruno-m-antunes-alb:
LGTM

cloudwebrtc (approved):

PR #1903, opened 2025-07-28
fix: Remove Bluetooth permission from AudioSwitch

Lucifer1091:
This commit upgrades the AudioSwitch library.

The BLUETOOTH permission is no longer needed for apps targeting Android 12 (API 31) and above. It has been replaced by more specific permissions such as BLUETOOTH_SCAN, BLUETOOTH_CONNECT, and BLUETOOTH_ADVERTISE.

This change removes the legacy BLUETOOTH permission from the manifest.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -53,7 +53,7 @@ android {
 
 dependencies {
     implementation 'io.github.webrtc-sdk:android:137.7151.01'
-    implementation 'com.github.davidliu:audioswitch:89582c47c9a04c62f90aa5e57251af4800a62c9a'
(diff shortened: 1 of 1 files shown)


bruno-m-antunes-alb:
@Lucifer1091 can you please open a PR on the AudioSwitch repo so we can have this on the oficial repo?

davidliu:
As it is, this change isn't really needed, since the original permission is limited to API 30 in the first place.

https://github.com/davidliu/audioswitch/blob/89582c47c9a04c62f90aa5e57251af4800a62c9a/audioswitch/src/main/AndroidManifest.xml#L4

PR #1904, opened 2025-08-03
small setVolume addition

mroccultus:
Paired with https://github.com/flutter-webrtc/webrtc-interface/pull/33
Tested on web with chrome, never tested other platforms but since already presented Helper class do it can't see any problems with that. I thought it's actually may be of use, can't imagine production ready app without this functionality present clear where you expect it to find.

### DIFF
--- a/lib/src/native/rtc_video_platform_view_controller.dart
+++ b/lib/src/native/rtc_video_platform_view_controller.dart
@@ -180,4 +180,18 @@ class RTCVideoPlatformViewController extends ValueNotifier<RTCVideoValue>
     }
     return true;
   }
+
+  @override
+  Future<void> setVolume(double value) async {
+    try {
+      if (_srcObject == null) {
+        throw Exception('Can\'t set volume: The MediaStream is null');
+      }
+      for(MediaStreamTrack track in  _srcObject!.getAudioTracks()) {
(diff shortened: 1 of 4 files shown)


cloudwebrtc:
I recommend extending these methods in `VideoRendererExtension` instead of modifying the interface. This can cause older versions of `flutter-webrtc` to fail to compile after the `webrtc_interface` is updated because they don't implement overrides.

https://github.com/flutter-webrtc/flutter-webrtc/blob/main/lib/src/video_renderer_extension.dart

```dart
abstract class AudioControl {
  Future<void> setVolume(double volume);
}
```

for 
native/rtc_video_renderer_impl.dart 
native/rtc_video_platform_view_controller.dart
web/rtc_video_renderer_impl.dart

```dart
import '../video_renderer_extension.dart' show AudioControl;

class RTCVideoRenderer extends ValueNotifier<RTCVideoValue>
    implements VideoRenderer, AudioControl {
  RTCVideoRenderer()
 ....... 
 
   @override
  Future<void> setVolume(double volume) async {
     ........
  }
}

```

mroccultus:
separating muted and volume control a bit strange, but ok

cloudwebrtc (approved):

PR #1700, opened 2024-11-05
Add sharedSingleton for iOS/Android.

cloudwebrtc:

Issue #959, opened 2022-05-24
Support Insertable Streams


cloudwebrtc:
## Goals:
Implement the Dart version of insertable streams
1, can be used for E2EE,
2, Audio, Video file recording, screenshot
3, Generate MediaStreamTrack from raw data, and process audio and video database through pipeline

## The issue that will be solved
https://github.com/flutter-webrtc/flutter-webrtc/issues/691
https://github.com/flutter-webrtc/flutter-webrtc/issues/361
https://github.com/flutter-webrtc/flutter-webrtc/issues/957
https://github.com/flutter-webrtc/flutter-webrtc/issues/300
https://github.com/flutter-webrtc/flutter-webrtc/issues/632

#### Implement the following interfaces.

DataView
```js
class DataView {
   buffer: ByteBuffer;
}
```

EncodedAudioChunk
```js
class EncodedAudioChunk {
     type: 'key' | 'delta';
     timestamp: int;
     duration: int;
     data: DataView;
     byteLength: int;
}
```
EncodedVideoChunk
```js
class EncodedVideoChunk {
     type: 'key' | 'delta';
     timestamp: int;
     duration: int;
     data: DataView;
     byteLength: int;
}
```

AudioFrame 
```js
class AudioFrame {
    format: 'u8' | 's16' | 's32';  //??
    data: DataView;
    timestamp: int;
    duration: int;
}
```

VideoFrame
```js
class VideoFrame {
   format: 'rgb' | 'yuv';  //??
   data: DataView;
   timestamp: int;
   duration: int;
}
```

WritableStream
```js
class WritableStream {
   write(chunk);  // EncodedAudioChunk |  EncodedVideoChunk | AudioFrame | VideoFrame
}
```

ReadableStream
```js
class ReadableStream {
 // return  EncodedAudioChunk |  EncodedVideoChunk | AudioFrame |VideoFrame
  Data read(); 
  pipeThrough(transform: TransformStream);
  pipeTo(destination: WritableStream);
  Array<ReadableStream> tee(); //copy out two streams
}
```

[MediaStreamTrackGenerator](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrackGenerator)
```js
class MediaStreamTrackGenerator {
    constructor({kind: 'audio' || 'video'});
    WritableStream writeable; // write AudioFrame or VideoFrame here.
};
```

[...]

PR #1920, opened 2025-08-26
feat!: make processing interface async

holzgeist:
:warning: BREAKING CHANGE

this PR changes the processor interface to be async. Instead of `VideoFrame onFrame(VideoFrame frame)`, which requires to process the frame immediately on the current thread, basing the interface on the [`VideoSink`](https://chromium.googlesource.com/external/webrtc/+/HEAD/sdk/android/api/org/webrtc/VideoSink.java) interface allows dispatching frame processing to thread pools, GPU etc.

The new interface requires two functions:
- `void setSink(VideoSink videoSink)` which provides a processor with the next processor or a final sink/renderer which is should forward processed frames to. If it is null, neither processing nor forwarding is needed
- `void onFrame(VideoFrame frame)` (from `VideoSink`), which tells the processor about a new frame. Depending on the current workload (FPS, previous frame still processing, etc), the processor can chose to ignore the frame or process it.

The change is done in a similar way on iOS

There might be a way to make this backwards compatible with older, synchronous implementations of the interface. If there is a chance that this is merged, I can look into it


M4dhav:
Hey @holzgeist , I have a question that's not necessarily related to this PR but instead to #1700 which implemented the Sink APIs, I wanted to know how exactly to access these APIs from the dart side as I couldn't find any documentation for the same. I figure you might have used them, so I was wondering if you could help me with this. I want to access the raw Audio Data coming from the server using this method to then pass it into an on device transcription service.

holzgeist:
Hi @M4dhav,

AFAIK they are not usable from the dart side (yet). The plan seems to be to expose dart FFI at some point (#959), but for now any processing needs to be implemented natively on each platform

M4dhav:
Hey, thanks a lot for the reply. If it's not too much trouble could you please send me some documentation on how i would go about implementing this platform wise?

holzgeist:
We are currently in the process of open sourcing our video blurring plugin here: https://github.com/holzgeist/flutter-livekit-background-effects
It's still in an early stage, so expect bugs and breaking changes. But it's in a usable state

PR #1925, opened 2025-09-03
upgrade compileSdk to 36 as standard for 16kb pages support

shivanshtalwar0:
Without this android app with ndk 28 isn't compiling and 16kb page support is made mandatory by google as it is supported from ndk 28 this minor fix makes app working again


cloudwebrtc:
@shivanshtalwar0 I think we need to revert the changes in `android/.gradle`, which don't seem to be necessary

shivanshtalwar0:
@cloudwebrtc  done!

cloudwebrtc (approved):

PR #1941, opened 2025-10-13
Fix audio session output port override for speaker toggle

md-riaz:
This pull request makes a targeted update to the `setSpeakerphoneOn:` method in `AudioUtils.m`, improving how the speakerphone audio route is overridden. The change replaces a deprecated or incorrect constant with the correct `AVAudioSessionPortOverrideSpeaker` value, ensuring compatibility and correctness.

- Audio routing update:
  * Updated the call to `overrideOutputAudioPort` to use `AVAudioSessionPortOverrideSpeaker` instead of `kAudioSessionProperty_OverrideAudioRoute` in the `setSpeakerphoneOn:` method in `AudioUtils.m`.

### DIFF
--- a/common/darwin/Classes/AudioUtils.m
+++ b/common/darwin/Classes/AudioUtils.m
@@ -93,7 +93,7 @@
                                         AVAudioSessionCategoryOptionAllowBluetooth
                                   error:&error];
 
-    success = [session overrideOutputAudioPort:kAudioSessionProperty_OverrideAudioRoute
+    success = [session overrideOutputAudioPort:AVAudioSessionPortOverrideSpeaker
                                          error:&error];
     if (!success)
       NSLog(@"setSpeakerphoneOn: Port override failed due to: %@", error);


md-riaz:
@cloudwebrtc please check this

md-riaz:
@cloudwebrtc can you check this please?

cloudwebrtc:
hey @md-riaz , I suspect that using AVAudioSessionPortOverrideSpeaker will force override when you connect Bluetooth or wired headphones. Flutter-webrtc by default expects setSpeakerOn to only turn on the speaker when no external headphones (Bluetooth or wired) are connected. In other cases, wired/Bluetooth headphones should have higher priority than the Speaker to better align with user habits.

md-riaz:
Hi @cloudwebrtc — thanks for the feedback and for maintaining this amazing library for the community. I really want flutter-webrtc to be the best it can be for every usage, so I appreciate the careful review.

I want to clarify that my change was only to correct a deprecated/incorrect constant and not to alter any intended behavior:

- What I changed: replaced the old AudioSession-era/deprecated value with AVAudioSessionPortOverrideSpeaker in the overrideOutputAudioPort call.
- Why: the previous constant was not correct for the AVAudioSession API; using AVAudioSessionPortOverrideSpeaker aligns the call with the current API and restores the intended semantic.
- Intended effect: I did not intend to change routing policy. The goal was to fix the API usage so the speaker toggle behaves as originally intended — i.e., when a user explicitly toggles the speaker on, the app is requesting speaker output.

Regarding your concern that this could force audio to the speaker when Bluetooth/wired headphones are connected: if a user explicitly chooses to enable the speaker, that is a user-driven decision to listen through the loudspeaker instead of the connected device. If setSpeakerOn does not produce that result in such cases, that would indicate the speaker-toggle functionality is broken in those scenarios.

At the same time, whether the speaker toggle should be blocked when external devices are connected is a higher-level integration decision. I believe the library should provide the correct and predictable API (a working setSpeakerOn) and allow app developers to implement their own policy (e.g., skip speaker override when an external output is connected) if they prefer. If integrators want that behavior, it should be handled in their app logic or via an opt-in helper, not enforced unconditionally by the library.

Options we can take next:
- Keep this PR as the API-correcting fix only (preferred for correctness).
- Add an optional guard/flag or helper that skips override when external outputs (Bluetooth/wired) are detected, and document the behavior so integrators can opt in.

Again — thank you for the review and for your work on this library.

md-riaz:
@cloudwebrtc I would like to have your opinion on the comment please 🤔

md-riaz:
@cloudwebrtc

cloudwebrtc:
Hi everyone， I will do some testing and get back to you.

stanislavurban-daktela:
@cloudwebrtc any updates?

md-riaz:
@cloudwebrtc

reduxdj:
any updates?

md-riaz:
@cloudwebrtc many people are waiting for this, please either merge this or introduce any other fix for this issue please

cloudwebrtc:
Sorry for the long delay. I just did a quick test and it seems there are no side effects.

PR #1969, opened 2025-12-09
fix:hisi chip Error 0x80001001

iBxsr:
Huawei Phone use MediaRecorder failed.
Exception:
android.media.MediaCodec$CodecException: Error 0x80001001

@cloudwebrtc  Please check!


cloudwebrtc (approved):
lg!

PR #1977, opened 2025-12-29
macOS: Refresh screen capture sources & permission check

hiroshihorie:
Adds macOS/Catalyst screen‑recording permission handling and forces a fresh desktop source list on reload to avoid blank screen‑share previews after reboot (Sequoia
  15+).


cloudwebrtc (approved):
nice!

PR #1990, opened 2026-01-29
fix(macos): use CoreAudio ADM to prevent screen share crash

eccentricyash:
## Summary
Switches from AVAudioEngine ADM to CoreAudio ADM on macOS to fix a crash when toggling the microphone during screen share.

## Problem
On macOS, when screen sharing with system audio, toggling the microphone causes this crash:

```
AVAudioIONodeImpl::SetOutputFormat: required condition is false: format.sampleRate == hwFormat.sampleRate
```

AVAudioEngine struggles to handle the format changes when both screen share audio and microphone audio are active simultaneously.

## Solution
Use CoreAudio ADM (value `0`) instead of `RTCAudioDeviceModuleTypeAudioEngine`. CoreAudio handles multiple audio sources without the format conflicts that cause the crash.

Also added a null check to `postEvent` since the eventSink can be nil during rapid screen share state transitions.

## Testing
- Start screen share with audio on macOS
- Toggle mic on/off during screen share
- Confirmed no crash and mic audio works correctly

### DIFF
--- a/common/darwin/Classes/FlutterWebRTCPlugin.h
+++ b/common/darwin/Classes/FlutterWebRTCPlugin.h
@@ -13,7 +13,7 @@
 @class FlutterRTCMediaRecorder;
 @class AudioManager;
 
-void postEvent(FlutterEventSink _Nonnull sink, id _Nullable event);
+void postEvent(FlutterEventSink _Nullable sink, id _Nullable event);
 
 typedef void (^CompletionHandler)(void);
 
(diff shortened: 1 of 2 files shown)


melio-matt:
@eccentricyash I upgraded my version of flutter_webrtc to 1.3.1 from 1.2.0 having originally come from 0.12.6. When using 1.2.0 I found there was around a 3 second delay before audio packets started to be outputted from my Mac. Not what you want when you answer the phone. I then updated to 1.3.1 and this problem was solved and I think the reason was for this change that you made. However the side effect is that I seem to have lost echo cancellation. On my original build with 0.12.6 it didn't have an echo problem, nor did I on version 1.2.0. I'm using a MacBook Air M2 and the internal microphone and speakers. Is there a way to turn the echo cancelation back on with CoreAudio?

PR #2019, opened 2026-03-19
fix(ios): stabilize WebRTC audio session recovery and output device routing

jmgiannattasio:
## Summary
This PR fixes iOS audio device/routing instability in our WebRTC call flow (especially around CallKit answer/background transitions), where calls could connect but still have no audio or route to the wrong output device.

## Problem
On iOS we were seeing intermittent audio failures such as:
- `RTCAudioSession` activation failures (`Session activation failed`)
- playout initialization failures (`InitPlayout` / `output: -1`)
- route/category transitions unexpectedly switching output (Speaker/Receiver)
- video calls sometimes ending up on earpiece instead of speaker

## Root Cause
The issue was a combination of:
1. Audio session transitions (CallKit ↔ WebRTC) leaving `RTCAudioSession` temporarily inactive.
2. WebRTC AudioDeviceModule (ADM) not always recovering playout/recording after those transitions.
3. Route/category changes reverting output selection during active calls.

### 1) iOS audio session recovery
- Added bounded recovery attempts when the audio session fails to become active.
- Triggered recovery on relevant iOS route/category changes while a call is alive.

### 2) WebRTC ADM recovery
- Added explicit ADM recovery to re-initialize/start playout and recording when needed.
- This prevents “connected call but no audio” states after session interruptions.

### 3) Explicit local audio restart path
- Added a native restart entrypoint (`restartLocalAudio`) to allow controlled audio pipeline restart from upper layers.

### 4) Safer output routing behavior on iOS
- Hardened Speaker/Receiver handling during route/category transitions.
- Ensures video-call routing expectations (speaker-first behavior) remain stable.

## Expected Impact
- Fewer iOS cases of “call connected but no audio”.
- More stable audio output routing across CallKit/category/route transitions.
- Improved reliability for both inbound and outbound WebRTC calls on iOS.

## Scope
- No intended behavioral changes for Android.

### DIFF
--- a/common/darwin/Classes/FlutterWebRTCPlugin.m
+++ b/common/darwin/Classes/FlutterWebRTCPlugin.m
@@ -22,6 +22,7 @@
 
 #import <WebRTC/RTCLogging.h>
 #import <WebRTC/RTCCallbackLogger.h>
+#import <stdarg.h>
 
 #if TARGET_OS_OSX || TARGET_OS_MACCATALYST
 #import <CoreGraphics/CoreGraphics.h>
@@ -115,6 +116,10 @@ void postEvent(FlutterEventSink _Nullable sink, id _Nullable event) {
   id _textures;
   BOOL _speakerOn;
   BOOL _speakerOnButPreferBluetooth;
+  BOOL _isRestoringRouteAfterCategoryChange;
+  BOOL _didRouteRecoveryForCurrentCall;
+  BOOL _audioSessionRecoveryPending;
+  BOOL _isRunningAudioSessionRecovery;
   AVAudioSessionPort _preferredInput;
   AudioManager* _audioManager;
 #if TARGET_OS_IPHONE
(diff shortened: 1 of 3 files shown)


common/darwin/Classes/FlutterWebRTCPlugin.m
```diff
@@ -848,6 +904,14 @@ - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result {
       }
       [dataChannels removeAllObjects];
     }
+    if (self.peerConnections.count == 0) {
+      _speakerOn = NO;
```
cloudwebrtc: QQ: When we establish a PC connection and then close it, do we lose the previous state of _speakerOn and _speakerOnButPreferBluetooth, and might it revert to speaker off when the user establishes a PC connection again?

jmgiannattasio: Good catch. In the current implementation, speaker preference is not persisted across peer-connection lifecycles.

- The plugin resets `_speakerOn` / `_speakerOnButPreferBluetooth` to `NO` when creating a new PC (`createPeerConnection`).
- It resets them again when the last PC is closed/disposed.
- `deactiveRtcAudioSession` also clears that state when there are no local audio tracks and no active PCs.

So if a user called `setSpeakerOn(true)` in a previous call, then starts a new PC without calling `setSpeakerOn/selectAudioOutput` again, the new PC starts from a neutral state (`speakerOn == NO`), and iOS default routing applies.

This is intentional to avoid stale route state leaking from one call to the next.

PR #2049, opened 2026-05-19
Fix handling of `KeyDerivationAlgorithm` option

Airyzz:
Currently on the flutter side, you can set a Key Derivation Algorithm, but the value is not actually passed to libwebrtc. This correctly parses the option.

Also tweaking the function `KeyDerivationAlgorithmFromInt` to properly return `libwebrtc::KeyDerivationAlgorithm` as opposed to the currently used `libwebrtc::FrameCryptorAlgorithm`


cloudwebrtc (approved):
lg!

PR #2055, opened 2026-05-22
Fix android screen capture rotation

EnterDevelopers:
- Resize the existing VirtualDisplay instead of releasing/recreating it on orientation changes.
- Advertise H264 decoder support with at least level 4.1 (profile-level-id ...29) to allow higher-resolution iOS screen streams on Android.


cloudwebrtc (approved):
lg!

PR #2068, opened 2026-06-02
fix(desktop): map echoCancellation/noiseSuppression/autoGainControl constraints to RTCAudioOptions

### LINKED ISSUE #2067: [Windows/Linux] echoCancellation/noiseSuppression/autoGainControl constraints have no effect — software AEC/NS/AGC always on
## Summary

Since 1.2.1 the software APM (AEC/NS/AGC) is unconditionally enabled on Linux and Windows. Setting `echoCancellation: false` (or `noiseSuppression: false`, `autoGainControl: false`) from Dart has no effect.

## Root cause

In `common/cpp/src/flutter_media_stream.cc`, `GetUserAudio()` parses audio constraints into an `RTCMediaConstraints` object but never uses it. `CreateAudioSource` is called without an `RTCAudioOptions` argument:

```cpp
// Before fix — RTCAudioOptions omitted, all fields default to true
scoped_refptr<RTCAudioSource> source =
    base_->factory_->CreateAudioSource("audio_input");
```

Additionally, `ParseMediaConstraints` only reads `mandatory` and `optional` sub-keys, silently dropping flat W3C keys (`echoCancellation`, `noiseSuppression`, `autoGainControl`) that most callers (including the LiveKit Flutter SDK) pass at the top level.

The `getSettings()` response compounds the issue by hardcoding `echoCancellation: true` regardless of the actual constraint input.

## Steps to reproduce

```dart
final stream = await navigator.mediaDevices.getUserMedia({
  'audio': {
    'echoCancellation': false,
    'noiseSuppression': false,
    'autoGainControl': false,
  },
  'video': false,
});
// On Linux/Windows: AEC/NS/AGC remain active. No audible difference.
```

## Expected behaviour

Setting `echoCancellation: false` disables the software echo canceller; `true` enables it (W3C default).

## Environment

- Platform: Linux (also affects Windows, eLinux — all share `common/cpp/`)
- flutter_webrtc: 1.2.1 – 1.4.1
- libwebrtc binary: m144.7559.02

alfredobs97:
Fixes #2067

## Problem

Since 1.2.1 the software APM (AEC/NS/AGC) is unconditionally enabled on Linux/Windows/eLinux. The `echoCancellation`, `noiseSuppression`, and `autoGainControl` constraint keys are parsed into an `RTCMediaConstraints` object that is then silently discarded. `CreateAudioSource` is called without an `RTCAudioOptions` argument, so all three processing stages default to `true` regardless of the Dart caller's intent.

`getSettings()` also hardcoded `true` for all three flags, masking the issue.

## Fix

- Add `getAudioProcessingFlag()`: resolves a boolean flag from all three constraint shapes — flat W3C key, `mandatory` sub-map, `optional` list entries (the format the LiveKit SDK and others use).
- Populate an `RTCAudioOptions` struct from the resolved values and pass it to `CreateAudioSource`.
- For the `audio: true` (boolean) path, keep all processing on to preserve existing behaviour.
- Report the actual constraint values in `getSettings()`.

Only `common/cpp/src/flutter_media_stream.cc` is changed — shared by Linux, Windows, and eLinux.

## Test procedure

1. On Linux (Ubuntu with PipeWire or PulseAudio), call `getUserMedia` with `echoCancellation: false`.
2. Play audio through speakers while recording — echo should be audible in the captured stream.
3. Repeat with `echoCancellation: true` — echo should be suppressed by the software AEC.
4. Verify `track.getSettings()['echoCancellation']` returns the value passed in the constraint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 
 # Changelog
 
+[Unreleased]
+
+* [Windows/Linux] fix: Map echoCancellation/noiseSuppression/autoGainControl constraints to RTCAudioOptions so software AEC/NS/AGC can actually be toggled from Dart (#XXXX).
+
 [1.4.1] -2026-03-24
 
 * [Dart] fixed scalabilityMode (#2022).
--- a/common/cpp/src/flutter_media_stream.cc
+++ b/common/cpp/src/flutter_media_stream.cc
@@ -90,6 +90,61 @@ void addDefaultAudioConstraints(
   audioConstraints->AddOptionalConstraint("googDAEchoCancellation", "true");
 }
 
+// Reads a boolean audio-processing flag from the audio constraint map.
+// Supports flat W3C keys, mandatory sub-map keys, and optional list entries
+// (each a single-pair map, the format used by the LiveKit SDK and others).
+// Accepts bool values or "true"/"false" strings. Falls back to defaultValue
+// when the key is absent (W3C default is true for AEC/NS/AGC).
+static bool getAudioProcessingFlag(const EncodableMap& audioMap,
+                                   const std::vector<std::string>& keys,
+                                   bool defaultValue) {
+  auto readBoolValue = [](const EncodableValue& v, bool def) -> bool {
+    if (TypeIs<bool>(v)) return GetValue<bool>(v);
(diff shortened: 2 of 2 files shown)


cloudwebrtc (approved):
nice!

PR #2079, opened 2026-06-12
feat(android): add fullScreenOnly option to requestCapturePermission (force entire-screen on API 34+)

### LINKED ISSUE #2080: [Android] Allow forcing entire-screen capture (remove single-app option) on Android 14+
## Feature request

On Android 14 (API 34+), the `MediaProjection` consent dialog presents a dropdown letting the user choose between **"Share one app"** and **"Share entire screen"**. For use cases that only make sense with full-screen capture — remote support, screen mirroring, kiosk/monitoring — the single-app option is confusing and unusable: if the user picks it, the captured stream is just one app, not the device screen the operator expects.

Android provides a way to remove that choice: passing `MediaProjectionConfig.createConfigForDefaultDisplay()` to `MediaProjectionManager.createScreenCaptureIntent(config)` forces entire-screen capture and the consent dialog no longer offers the single-app option.

Currently `flutter_webrtc` always calls the no-arg `createScreenCaptureIntent()`, so apps cannot opt into full-screen-only behavior.

## Proposal

Add an opt-in flag to `Helper.requestCapturePermission()`:

```dart
// Android 14+: only offer entire-screen capture in the consent dialog
await Helper.requestCapturePermission(fullScreenOnly: true);
```

- Opt-in, defaults to `false` → existing behavior unchanged.
- API-gated: only applies the config on API 34+; older versions fall back to the current `createScreenCaptureIntent()`.
- No effect on macOS.

## Environment

- Affects Android 14 (API 34) and above.

## Implementation

Proposed in #2079 (verified on an Android 16 / API 36 emulator: with `fullScreenOnly: true` the dialog defaults to entire-screen and the single-app option is disabled; with the default `false` the user-choice dialog is unchanged).

deviankim:
Resolves #2080

## Summary

On Android 14 (API 34+), the `MediaProjection` consent dialog lets the user choose between sharing a **single app** or the **entire screen**. Some apps (remote support, screen mirroring) only support full-screen capture and the single-app option is confusing/unusable for them.

This adds an opt-in `fullScreenOnly` flag to `Helper.requestCapturePermission()`. When `true` and running on API 34+, the screen-capture intent is created with `MediaProjectionConfig.createConfigForDefaultDisplay()`, which forces entire-screen capture and removes the single-app option from the consent dialog.

```dart
// only offer entire-screen capture on Android 14+
await Helper.requestCapturePermission(fullScreenOnly: true);
```

## Details

- **Opt-in, non-breaking**: defaults to `false`, so existing callers and the `getDisplayMedia` path are unchanged.
- **API-gated**: the `MediaProjectionConfig` branch only runs on `Build.VERSION_CODES.UPSIDE_DOWN_CAKE` (API 34) and above; older versions fall back to the existing `createScreenCaptureIntent()`.
- **No effect on macOS**: the flag is threaded only through the Android `MediaProjection` path.
- Flag flows: `helper.dart` → `MethodCallHandlerImpl` (`fullScreenOnly` arg, null-safe) → `GetUserMediaImpl.requestCapturePermission` → `screenRequestPermissions` (a backward-compatible 1-arg overload delegates to `false`) → Fragment args → `requestStart`.

## Testing

Built and verified on an Android 16 (API 36) emulator:
- `fullScreenOnly: true` → consent dialog defaults to "entire screen" and the single-app option is disabled ("not supported by this app").
- `fullScreenOnly: false` (default) → unchanged user-choice dialog.

### DIFF
--- a/android/src/main/java/com/cloudwebrtc/webrtc/GetUserMediaImpl.java
+++ b/android/src/main/java/com/cloudwebrtc/webrtc/GetUserMediaImpl.java
@@ -13,6 +13,7 @@ import android.graphics.Point;
 import android.hardware.camera2.CameraManager;
 import android.media.AudioDeviceInfo;
 import android.media.projection.MediaProjection;
+import android.media.projection.MediaProjectionConfig;
 import android.media.projection.MediaProjectionManager;
 import android.net.Uri;
 import android.os.Build;
@@ -100,6 +101,7 @@ public class GetUserMediaImpl {
     private static final String PROJECTION_DATA = "PROJECTION_DATA";
     private static final String RESULT_RECEIVER = "RESULT_RECEIVER";
     private static final String REQUEST_CODE = "REQUEST_CODE";
+    private static final String FULL_SCREEN_ONLY = "FULL_SCREEN_ONLY";
 
     static final String TAG = FlutterWebRTCPlugin.TAG;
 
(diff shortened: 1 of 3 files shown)


cloudwebrtc (approved):
lg

PR #2090, opened 2026-06-16
fix(macos): create parent directory before captureFrame writes the file

### LINKED ISSUE #2089: macOS/iOS: captureFrame() fails when the destination directory does not exist
### Description

On Apple platforms the shared `common/darwin/Classes/FlutterRTCFrameCapturer.m` writes the captured frame straight to disk with `[imageData writeToFile:_path atomically:NO]`. If the parent directory of the requested path does not already exist, the write fails and the method returns a generic failure — the capturer never attempts to create the directory.

### Steps to reproduce

1. Call `MediaStreamTrack.captureFrame(path)` with a path whose parent directory has not been created yet, e.g. `<tmpDir>/snapshots/frame.jpg` where `snapshots/` does not exist.
2. The write fails; the caller receives an opaque failure with little context.

### Expected

`captureFrame(path)` should create the intermediate directories before writing (the intuitive expectation for a fresh path) and surface a clear error if the write still fails.

### Proposed fix

Create the parent directory via `NSFileManager createDirectoryAtPath:withIntermediateDirectories:` before the write, and make the failure log actionable (whether `imageData` was nil + the target path). A PR implementing this follows.

### Environment

- Platforms: macOS / iOS (shared darwin capturer)
- flutter_webrtc: v1.5.1 / `main`

deviankim:
### What

`FlutterRTCFrameCapturer` (shared by iOS/macOS) writes the captured frame with `[imageData writeToFile:_path atomically:NO]`. When the parent directory of `_path` does not exist the write fails with only a generic `Failed to write to file` log.

This change creates the intermediate directories with `NSFileManager` before the write, and makes the failure path actionable by returning a `FlutterError` and logging whether `imageData` was nil plus the target path.

### Why

Closes #2089. Callers that pass a not-yet-created subdirectory currently get an opaque failure; creating the directory first matches the expectation that `captureFrame(path)` works for a fresh path.

### Changes

- `common/darwin/Classes/FlutterRTCFrameCapturer.m`: create parent directory before writing; clearer error/log on failure.

### Notes

- Single file, no public API change.
- Original change authored by 송건호 <ghsong@rsupport.com>; submitted via fork.

### DIFF
--- a/common/darwin/Classes/FlutterRTCFrameCapturer.m
+++ b/common/darwin/Classes/FlutterRTCFrameCapturer.m
@@ -93,11 +93,26 @@
   CGImageRelease(cgImage);
   if (shouldRelease)
     CVPixelBufferRelease(pixelBufferRef);
+
+  // Ensure the parent directory exists before writing the file
+  NSString* parentDir = [_path stringByDeletingLastPathComponent];
+  NSFileManager* fileManager = [NSFileManager defaultManager];
+  if (![fileManager fileExistsAtPath:parentDir]) {
+    NSError* dirError = nil;
+    [fileManager createDirectoryAtPath:parentDir
+           withIntermediateDirectories:YES
+                            attributes:nil
+                                 error:&dirError];
+    if (dirError) {
+      NSLog(@"[CaptureFrame] Failed to create directory: %@, error: %@", parentDir, dirError);
+    }
+  }
+
   if (imageData && [imageData writeToFile:_path atomically:NO]) {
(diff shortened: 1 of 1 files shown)


hiroshihorie (approved):
LGTM. Verified against main, the failure path already returned a FlutterError so this is purely the parent directory creation plus better failure logging, both welcome. The fileExistsAtPath pre-check is redundant since createDirectory with intermediates succeeds when the directory exists, but harmless.

PR #2062, opened 2026-05-26
feat(ios,macos): add Swift Package Manager integration.

cloudwebrtc:
Restructure plugin sources from Classes/ into the Flutter SPM convention
flutter_webrtc/Sources/flutter_webrtc/ with public headers under
include/flutter_webrtc/. Source symlinks into common/darwin/Classes are
preserved so it remains the single source of truth.

- Add ios/flutter_webrtc/Package.swift and macos/flutter_webrtc/Package.swift,
  both pulling WebRTC 144.7559.04 directly via .binaryTarget against the
  xcframework release asset (bypasses webrtc-sdk/Specs Package.swift, which
  declares .visionOS(.v26) under swift-tools 5.9 and fails to compile).
- Update both podspecs to point at the new Sources/ layout and add
  USER_HEADER_SEARCH_PATHS so quoted #import "Foo.h" keeps resolving.
- Drop the broken stub macos/Package.swift in favor of the new location.
- Example: deintegrate CocoaPods from macOS (all macOS deps are SPM-ready)
  and from iOS after removing permission_handler and gallery_saver_plus
  from pubspec and the two samples that referenced them, since neither
  ships a Package.swift upstream.

Verified with Flutter 3.41 + Xcode 26.5: flutter build macos and an
xcodebuild iphonesimulator build of the example both succeed.

close #2045, close #2027

---

## Update 2026-07-29 (takeover by @hiroshihorie)

Changes since the original description:

- Caught up with main (twice), no conflicts beyond CHANGELOG.
- Bumped the Package.swift binary targets from WebRTC 144.7559.04 to 144.7559.09 to match the podspecs. The mute mode APIs merged from main only exist in .09, so the old pin no longer compiled. This also fixes the `isPlatformVoiceProcessingAllowed` error reported in the comments.
- Exported the WebRTC binary target as a library product so dependent plugins (livekit_client, see livekit/client-sdk-flutter#1142) can reuse it instead of declaring a colliding second copy.
- Restored `permission_handler` to the example. permission_handler_apple 9.4.10 now ships a Package.swift, so it no longer blocks the CocoaPods free example. `gallery_saver_plus` stays removed (no SPM support upstream), a replacement using `gal` can land separately.
- Removed the leftover `ios/Classes` and `macos/Classes` platform view symlinks that came back through merges. The podspecs point at the `Sources/` layout, so they were dead.
- Synced the podspec versions with pubspec (1.5.2).
[...]

PR #2092, opened 2026-06-16
fix(android): resolve OrientationAwareScreenCapturer deadlock/ANR on stop

### LINKED ISSUE #2091: Android: ANR/deadlock when stopping screen capture (synchronized stopCapture re-enters the capturer monitor)
### Description

`OrientationAwareScreenCapturer.stopCapture()` is declared `public synchronized`, so it holds the capturer monitor while it calls `ThreadUtils.invokeAtFrontUninterruptibly(...)` and blocks on the `SurfaceTextureHelper` thread. That same thread runs `onFrame()` → `changeCaptureFormat()`, which is **also** `synchronized` and therefore tries to re-enter the monitor that `stopCapture()` is still holding. The two block each other → deadlock → ANR while stopping screen capture.

### Where (current `main`)

- `stopCapture()` is `synchronized` and waits on the SurfaceTextureHelper thread.
- `onFrame()` calls `changeCaptureFormat()` on every frame, and `changeCaptureFormat()` is `synchronized`.
- `isDisposed` is non-volatile and there is no "stopped" guard, so in-flight frames keep entering the lock during stop.

PRs #2048 / #2055 reworked orientation/rotation correctness in this file but did not address this stop-time deadlock.

### Expected

Stopping screen capture must not be able to deadlock against the frame-delivery thread.

### Proposed fix

- Drop `synchronized` from `stopCapture()`; guard with a `volatile` stopped/disposed flag and return early.
- In `onFrame()`, only enter `changeCaptureFormat()` when the dimensions actually change (compare against `volatile` old width/height), which removes the per-frame lock and shrinks the race window.
- Make the relevant flags `volatile`; silently drop frames that arrive after stop/dispose.

A PR implementing this follows.

### Environment

- Platform: Android (screen capture / `OrientationAwareScreenCapturer`)
- flutter_webrtc: v1.5.1 / `main`

deviankim:
### What

Removes a stop-time deadlock in `OrientationAwareScreenCapturer`.

`stopCapture()` was `public synchronized`, so it held the capturer monitor while blocking on the `SurfaceTextureHelper` thread via `invokeAtFrontUninterruptibly`. That thread runs `onFrame()` → `changeCaptureFormat()` (also `synchronized`), which then re-enters the monitor `stopCapture()` still holds → deadlock → ANR.

### Why

Closes #2091. PRs #2048 / #2055 fixed orientation/rotation correctness in this file but left the stop-time deadlock unaddressed.

### Changes

- Drop `synchronized` from `stopCapture()`; guard with a `volatile isStopped`/`isDisposed` flag and return early.
- `onFrame()` only enters `changeCaptureFormat()` when the target dimensions actually change (compared against `volatile` old width/height), removing the per-frame lock and shrinking the race window.
- Make `oldWidth`/`oldHeight`/`isDisposed` `volatile`, add `volatile isStopped`, and silently drop frames that arrive after stop/dispose.

### Notes

- Single file (`OrientationAwareScreenCapturer.java`).
- Original change authored by 오세진 <ohsj@rsupport.com>; submitted via fork.


deviankim:
### Windows CI failure appears unrelated to this PR (libwebrtc extraction in `third_party/CMakeLists.txt`)

The only failing check is **Build for flutter Windows** — Android, iOS, macOS, Web, Linux, Embedded Linux and Dart Format all pass. The Windows job fails inside the libwebrtc download/extract step, not in this changeset:

```
CMake Error: Problem with archive_write_header(): Cannot extract through symlink
  ...\example\windows\flutter\ephemeral\.plugin_symlinks\flutter_webrtc
CMake Error at .../third_party/CMakeLists.txt:89 (file): file failed to extract
  .../third_party/CMakeLists.txt:128 (_libwebrtc_extract_and_normalize)
```

Evidence that this is a Windows-runner / `third_party` extraction infra issue rather than this PR:
- This PR does not touch `third_party/CMakeLists.txt` or any Windows build files.
- An earlier run of this branch **passed** the Windows build; re-running now reproduces the same `Cannot extract through symlink` error (so it is not flaky-in-my-favor — it reproduces deterministically after an infra change).
- The **same** failure currently appears on unrelated PR #2062.

Note: this PR only changes a few code comments in one Java file, yet the Windows build fails identically to other PRs — further indication the failure is environmental, not code.

Could a maintainer confirm whether this is a known Windows libwebrtc-extraction issue and re-run the Windows job once resolved? Happy to rebase if needed. Thanks!

hiroshihorie (approved):
The deadlock analysis is correct and the fix is sound: dropping the monitor from stopCapture breaks the stop-vs-onFrame circular wait, the unlocked dimension check is safely re-verified inside the synchronized changeCaptureFormat, and silently dropping post-stop frames beats the old throw on the render thread. One non blocking note: isStopped is never reset in startCapture, so a stopped instance can't restart. Unreachable through this plugin (screen capturers are single use) but a one liner would keep the general VideoCapturer contract since the class is public. The Windows CI failure predates #2100 and is unrelated, updating the branch to re-run.

PR #2099, opened 2026-06-19
Expose Android audio device module

hiroshihorie:
## Summary

Expose the Android `JavaAudioDeviceModule` from `FlutterWebRTCPlugin` through a narrow getter.

## Why

Downstream plugins that already coordinate with `FlutterWebRTCPlugin.sharedSingleton` need access to the same ADM instance that flutter-webrtc creates and passes into `PeerConnectionFactory`. This lets them start/prewarm recording with options without duplicating or replacing flutter-webrtc's ADM.

## Impact

This is an Android-only API surface for native plugin integrations. It does not change existing Dart APIs or existing start/stop recording behavior.

## Validation

- `git diff --check`


hiroshihorie:
@copilot Are your findings addressed with the new commit ?

PR #2107, opened 2026-07-08
Feat: Custom LocalVideoTrack - with this changes we can create a videoTrack on the native side of ios and android and attach it to a mediaStream

baldarama:
Use case:
sending a static image or pre recorded video (no audio yet) though a webrtc connection
sending a static image when the OS cuts access to the camera due to the app being minimized

(Android) FlutterWebRTCPlugin.sharedInstance.setVideoTrackFactory()
(iOS) FlutterWebRTCPlugin.sharedInstance.videoTrackFactory
by setting a video track factory, the flutter plugin can later obtain a customize videoTrack and add it to a mediaStream
videoTrackFactory can provide different custom VideoTrack based on a name

factory_impl.dart createLocalMediaStreamWithCustomVideoTrack()
calls a method channel that will create a local MediaStream and attach a custom VideoTrack to it, the custom VideoTrack will be generated by videoTrackFactory


baldarama:
i realize there is a way to accomplish what im trying to do with way less changes to the flutter-webrtc library, ill submit another or for that.

PR #2109, opened 2026-07-10
exposing public MediaStream getStreamForId in FlutterWebRTCPlugin.java

baldarama:
the same method is already exposed on the iOS implementation. i use this method to inject a custom video track in to the MediaStream to be able to send it through the webrtc connection.


hiroshihorie (approved):
LGTM, matches the existing embedder accessor pattern (getTrackForId, getLocalTrack). One note for callers: the underlying implementation NPEs if peerConnectionId is null, pass an empty string to search across all peer connections (getTrackForId tolerates null, so the two are slightly asymmetric, pre-existing and not this PR's fault).

PR #2115, opened 2026-07-20
feat(linux): system-audio (loopback) capture via PulseAudio

chefboyrdave21:
## What

Implements the Linux system-audio (loopback) capturer so
`getDisplayMedia({ video: true, audio: true })` streams system/desktop audio on
Linux, the same way it already works on Windows (WASAPI) and macOS.

Before this change, `CreateLoopbackCapturer()` on Linux was a `nullptr` stub, so
a Linux screen share silently carried no system-audio track.

## How

- **`linux/pulse_loopback_capturer.{h,cc}` (new):** uses the PulseAudio *simple*
  API to record the default output sink's `.monitor` source. The default sink is
  resolved through a short-lived `pa_context` introspection
  (`server_info` -> `sink_info.monitor_source_name`, falling back to
  `<sink>.monitor`). A single capture thread then block-reads
  S16LE / 48 kHz / stereo frames and feeds one 10 ms frame
  (`480 * 2ch * 2 bytes = 1920` bytes) at a time to
  `RTCAudioSource::CaptureFrame(..., 16, 48000, 2, 480)`, mirroring the Windows
  capturer. The blocking read self-paces to real time, so no ring buffer is
  needed.
- **`linux/loopback_capturer_factory.cc`:** returns the Pulse capturer under
  `HAVE_LIBPULSE`; otherwise keeps the previous `nullptr` behaviour, so the
  factory symbol is always defined for `__linux__`.
- **`linux/CMakeLists.txt`:** a guarded `pkg_check_modules(libpulse
  libpulse-simple)` enables the capturer when the dev headers are present, and
  degrades to the `nullptr` stub (build still succeeds) when they are not.

### Robustness

The introspection is bounded by a 3-second steady-clock deadline and runs the
mainloop non-blocking, so a misbehaving PulseAudio/PipeWire daemon degrades to
"no audio track" instead of blocking the Flutter platform thread inside
`getDisplayMedia`.

### PipeWire

PipeWire ships a PulseAudio compatibility layer, so the same `libpulse` path
covers both the PulseAudio and PipeWire stacks.

### GCC 16 note

The CMakeLists also force-includes `<cstdint>` for the plugin target under
GCC/Clang, because the bundled libwebrtc headers (e.g. `rtc_types.h`) use
fixed-width integer types without including `<cstdint>`, which fails to compile
on GCC 14+/16. Happy to split this into its own commit/PR if preferred.

## Testing

Built and verified on Manjaro Linux (PipeWire), Flutter 3.41.2, GCC 16:
- plugin links `libpulse` / `libpulse-simple`, `HAVE_LIBPULSE` defined;
- a screen share with audio produces a PulseAudio source-output bound to the
  sink `.monitor` (not the microphone);
- graceful degradation confirmed when `libpulse` dev headers are absent.

## Compatibility

No API changes. No behaviour change on Windows/macOS. On Linux without
`libpulse`, behaviour is identical to today (no audio track).

### DIFF
--- a/linux/CMakeLists.txt
+++ b/linux/CMakeLists.txt
@@ -29,6 +29,7 @@ add_library(${PLUGIN_NAME} SHARED
   "../common/cpp/src/flutter_webrtc_base.cc"
   "../common/cpp/src/flutter_common.cc"
   "loopback_capturer_factory.cc"
+  "pulse_loopback_capturer.cc"
   "flutter_webrtc_plugin.cc"
   "flutter/core_implementations.cc"
   "flutter/standard_codec.cc"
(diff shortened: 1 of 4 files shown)


cloudwebrtc (approved):
nice PR.

PR #2125, opened 2026-07-29
fix(android): refresh remote video tracks after renegotiation

### LINKED ISSUE #2124: Android RTCVideoView remains black after Cloudflare SFU renegotiation
**Describe the bug**

When connecting on Android to Cloudflare Realtime Serverless SFU, the WebRTC negotiation and subsequent subscription renegotiation complete successfully, but `RTCVideoView` remains black and never renders the first frame.

Inbound RTP statistics show that video packets and decoded frames are being received.

During the Unified Plan renegotiation, a replacement Java `MediaStream`/`VideoTrack` object is exposed with the same logical track ID. The Android renderer remains attached to the previous track object, which no longer receives frames.

**Expected behavior**

The renderer should automatically bind to the current remote `VideoTrack` after renegotiation and display the decoded video.

**Platform information**

- Platform: Android
- Plugin version: 1.5.2
- Service: Cloudflare Realtime Serverless SFU

vemax78:
Fixes #2124

## Summary

Fixes an Android video rendering issue observed when connecting to Cloudflare Realtime Serverless SFU.

After the SFU subscription renegotiation completed successfully, the remote video remained black and `RTCVideoView` never rendered its first frame, even though inbound RTP statistics showed that video frames were being received and decoded.

## Root cause

During a Unified Plan renegotiation, a new Java `MediaStream`/`VideoTrack` wrapper can be exposed while preserving the same logical WebRTC track ID.

The Android renderer remained attached to the previous `VideoTrack` object. As a result, the active remote track received decoded frames while the `VideoSink` used by `FlutterRTCVideoRenderer` was still connected to the stale track wrapper.

A late removal of the previous stream could also remove the replacement track from the track lookup because both objects shared the same track ID.

## Solution

This change:

- introduces a thread-safe registry containing the current remote track object for each track ID;
- updates the registry whenever a remote stream exposes a replacement track;
- automatically rebinds matching Android video renderers on the main thread;
- removes tracks using object identity, preventing a late removal event for an old wrapper from removing its replacement;
- allows the renderer to bind directly to the resolved `VideoTrack`;
- stores stream and track IDs for lifecycle bookkeeping instead of retaining potentially disposed `MediaStream` wrappers;
- safely detaches the previous `VideoSink` during track replacement and renderer disposal.

## Result

After renegotiation, the renderer is automatically moved to the current remote `VideoTrack`. The first frame is rendered correctly and the video continues playing instead of leaving a black `RTCVideoView`.

## Tests

- Added a unit test covering replacement tracks with the same ID and late removal of the previous track object.
- Android `testDebugUnitTest` passes.
- Dart formatting and Flutter analysis pass.
- Flutter tests pass.
- Repeated manual connections through Cloudflare Realtime Serverless SFU render video correctly and complete teardown without exceptions.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -78,4 +78,5 @@ dependencies {
     implementation 'io.github.webrtc-sdk:android:144.7559.09'
     implementation 'com.github.davidliu:audioswitch:039a35aefab7747c557242fa216c9ea11743b604'
     implementation 'androidx.annotation:annotation:1.1.0'
+    testImplementation 'junit:junit:4.13.2'
 }
--- a/android/src/main/java/com/cloudwebrtc/webrtc/FlutterRTCVideoRenderer.java
+++ b/android/src/main/java/com/cloudwebrtc/webrtc/FlutterRTCVideoRenderer.java
@@ -23,12 +23,12 @@ public class FlutterRTCVideoRenderer implements EventChannel.StreamHandler {
     private static final String TAG = FlutterWebRTCPlugin.TAG;
     private final TextureRegistry.SurfaceProducer producer;
     private int id = -1;
-    private MediaStream mediaStream;
+    private String mediaStreamId;
 
     private String ownerTag;
 
     public void Dispose() {
-        //destroy
+        setVideoTrack(null);
         if (surfaceTextureRenderer != null) {
             surfaceTextureRenderer.release();
         }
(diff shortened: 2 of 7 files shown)


cloudwebrtc (approved):
lgtm!

cc @hiroshihorie,  Perhaps the LK SDK will also encounter the same issue?

vemax78:
I focused on Android because that is the platform currently used in my application, where I reproduced and verified the issue.  

A similar problem may also be possible on iOS, but I have not investigated or tested that platform, so I cannot confirm it. It may be worth validating separately.

PR #1795, opened 2025-03-22
fix: Try to fix the non-platform thread call error.

cloudwebrtc:

PR #2130, opened 2026-07-29
fix(linux): marshal MethodResultProxy callbacks to platform thread

### LINKED ISSUE #2126: Linux/GTK: method-call results from libwebrtc worker threads are silently dropped (SDP answer never returned)
**Describe the bug**

On Linux (GTK embedding), method-call *results* returned from libwebrtc worker threads are silently dropped, so Dart `Future`s for operations like `createAnswer` / `setRemoteDescription` never resolve. The most visible symptom: a subscriber's SDP renegotiation never completes — the Dart side awaits an answer that was computed but never delivered, so `livekit_client` never sends the SDP answer back to the server and the peer connection stalls.

Root cause: `MethodResultProxyImpl` (`common/cpp/src/flutter_common.cc`) forwards `Success`/`Error`/`NotImplemented` directly to Flutter's `MethodResult` regardless of the calling thread. On Linux/GTK, Flutter's method channel only delivers responses that are posted on the **platform thread**; responses from a worker/signaling thread are dropped without error. The sibling `EventChannelProxy` already solves this by marshalling through `TaskRunner` (added in #1795 / `1c2ff6f`), but that marshalling was never extended to `MethodResultProxy`.

`MethodResultProxy::Create` currently takes only `method_result`; the Linux plugin calls it without a `TaskRunner`:
```cpp
MethodResultProxy::Create(std::move(result))   // linux/flutter_webrtc_plugin.cc
```

**To Reproduce**

1. Linux desktop (GTK), `flutter-webrtc` latest `main`.
2. Establish a peer connection that requires the local side to *answer* (e.g. a LiveKit subscriber subscribing to a remote publisher), where libwebrtc invokes `createAnswer`/`setLocalDescription` callbacks on a worker thread.
3. Observe the Dart `Future` for the answer never completes; the remote side never receives the SDP answer; the connection hangs. No exception is thrown — the response is simply lost.

**Expected behavior**

`MethodResultProxy` callbacks are marshalled onto the platform thread (via the existing `TaskRunner`) before being delivered to Flutter, mirroring `EventChannelProxy`. The Dart `Future` resolves and the SDP answer is sent.

**Platform information**

* **OS:** Linux (Desktop / GTK)
* **OS version:** Fedora 43

**Proposed fix**

Add an optional `TaskRunner*` parameter to `MethodResultProxy::Create` (default `nullptr` for backward compatibility with other platforms). When set, `Success`/`Error`/`NotImplemented` enqueue the call onto the platform thread via `TaskRunner::EnqueueTask`, converting the `unique_ptr<MethodResult>` to `shared_ptr` for lambda capture and adding a null guard for double-call safety. The Linux plugin passes its `task_runner()`; other platforms keep the current direct-call path.

This applies the exact pattern already accepted upstream for `EventChannelProxy` (#1795).

---

Quantumheart:
fix(linux): marshal MethodResultProxy callbacks to platform thread (#2126)

**Problem**

libwebrtc invokes PeerConnection callbacks (`setRemoteDescription`,
`createAnswer`, etc.) on worker threads. `MethodResultProxyImpl` forwarded
these directly to Flutter's `MethodResult`, which on Linux/GTK silently
drops responses that arrive on a non-platform thread. As a result the Dart
`Future` for the operation never resolved, so subscriber SDP renegotiation
never completed — `livekit_client` never sent the SDP answer back to the
server and the peer connection stalled.

The sibling `EventChannelProxy` already solves this by marshalling through
`TaskRunner` (#1795 / `1c2ff6f`), but that marshalling was never extended to
`MethodResultProxy`.

**Fix**

Add an optional `TaskRunner*` parameter to `MethodResultProxy::Create`
(default `nullptr` so other platforms keep the current direct-call path).
When set, `Success`/`Error`/`NotImplemented` enqueue the call onto the
platform thread via `TaskRunner::EnqueueTask`. The `unique_ptr<MethodResult>`
is converted to `shared_ptr` for lambda capture (`std::function` requires
copyability), and a null guard guards against double-call since the proxy is
moved out on the first terminal call.

The Linux plugin passes its `task_runner()`; elinux and windows continue to
call `Create` without a runner (default `nullptr`).

### Testing

  `livekit_client`): subscriber SDP renegotiation completes and the answer is
  delivered where it previously hung on Linux/GTK.

Fixes #2126

### DIFF
--- a/common/cpp/include/flutter_common.h
+++ b/common/cpp/include/flutter_common.h
@@ -150,7 +150,8 @@ class MethodCallProxy {
 class MethodResultProxy {
  public:
   static std::unique_ptr<MethodResultProxy> Create(
-      std::unique_ptr<MethodResult> method_result);
+      std::unique_ptr<MethodResult> method_result,
+      TaskRunner* task_runner = nullptr);
 
   virtual ~MethodResultProxy() = default;
 
--- a/common/cpp/src/flutter_common.cc
+++ b/common/cpp/src/flutter_common.cc
@@ -32,40 +32,84 @@ std::unique_ptr<MethodCallProxy> MethodCallProxy::Create(
 
 class MethodResultProxyImpl : public MethodResultProxy {
  public:
-  explicit MethodResultProxyImpl(std::unique_ptr<MethodResult> method_result)
-      : method_result_(std::move(method_result)) {}
+  MethodResultProxyImpl(std::unique_ptr<MethodResult> method_result,
+                        TaskRunner* task_runner)
+      : method_result_(std::move(method_result)), task_runner_(task_runner) {}
   ~MethodResultProxyImpl() {}
 
-  // Reports success with no result.
-  void Success() override { method_result_->Success(); }
-
-  // Reports success with a result.
-  void Success(const EncodableValue& result) override {
-    method_result_->Success(result);
+  void Success() override {
+    if (!method_result_)
+      return;
+    if (task_runner_) {
+      std::shared_ptr<MethodResult> result(std::move(method_result_));
+      task_runner_->EnqueueTask([result]() { result->Success(); });
+    } else {
+      method_result_->Success();
+    }
   }
 
-  // Reports an error.
+  void Success(const EncodableValue& val) override {
+    if (!method_result_)
+      return;
+    if (task_runner_) {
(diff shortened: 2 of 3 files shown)


cloudwebrtc (approved):
lg!

PR #2131, opened 2026-07-29
fix: prevent event channel use-after-free on peer connection cleanup

### LINKED ISSUE #2127: Event channel use-after-free on peer connection close (observer destroyed before Dart cancels subscription)
**Describe the bug**

During peer connection teardown, the event-channel observer is destroyed too early, leading to a use-after-free / dangling callback when Dart still holds an active subscription to the connection's event channel.

In `common/cpp/src/flutter_peerconnection.cc`, `RTCPeerConnectionClose` erases the entry from `peerconnection_observers_` immediately:
```cpp
void FlutterPeerConnection::RTCPeerConnectionClose(...) {
  ...
  auto it = base_->peerconnection_observers_.find(uuid);
  if (it != base_->peerconnection_observers_.end())
    base_->peerconnection_observers_.erase(it);
  result->Success();
}
```
Dart's `RTCPeerConnection` keeps its EventChannel subscription alive until `dispose()` is called (which invokes `RTCPeerConnectionDispose`). Because the observer — and thus the event channel handler — is already gone by `Close`, any event the WebRTC stack emits between `Close` and `Dispose` (or Dart's cancel) hits a destroyed handler. Additionally, `~EventChannelProxyImpl` is currently empty, so even after the proxy is destroyed the stream handler can remain registered, leaving dangling callbacks.

**To Reproduce**

1. Any desktop platform; Linux GTK reproduces readily.
2. Open a peer connection, subscribe to its event channel on the Dart side.
3. Call `close()` then asynchronously `dispose()` (or swipe-kill / rapidly tear down a call) so there is a window where the WebRTC thread emits an ICE/state event after `Close` but before `Dispose`.
4. Observe a crash (use-after-free) or a callback into freed memory from the event channel.

**Expected behavior**

The event channel and its observer stay alive until `RTCPeerConnectionDispose` (when Dart has cancelled its subscription), and the `EventChannelProxy` destructor unregisters its stream handler so no dangling callbacks can fire after the proxy is gone.

**Platform information**

* **OS:** Linux (Desktop / GTK) — also affects other desktop platforms
* **OS version:** Fedora 43

**Proposed fix**

1. Move the `peerconnection_observers_.erase(...)` from `RTCPeerConnectionClose` to `RTCPeerConnectionDispose`, so the channel survives `Close`.
2. In `~EventChannelProxyImpl` (`common/cpp/src/flutter_common.cc`), call `channel_->SetStreamHandler(nullptr)` to drop the handler on destruction.

---

Quantumheart:
fix: prevent event channel use-after-free on peer connection cleanup (#2127)

**Problem**

During peer connection teardown, `RTCPeerConnectionClose` erases the entry
from `peerconnection_observers_` immediately, but Dart's `RTCPeerConnection`
keeps its EventChannel subscription alive until `dispose()` (which invokes
`RTCPeerConnectionDispose`). Any event the WebRTC stack emits between `Close`
and `Dispose` therefore hits a destroyed observer — a use-after-free.

Additionally, `~EventChannelProxyImpl` was empty, so the stream handler could
remain registered after the proxy was destroyed, leaving dangling callbacks.

**Fix**

- Move the `peerconnection_observers_.erase(...)` from `RTCPeerConnectionClose`
  to `RTCPeerConnectionDispose`, so the event channel survives `Close` (until
  Dart has cancelled its subscription).
- In `~EventChannelProxyImpl`, call `channel_->SetStreamHandler(nullptr)` to
  drop the handler on destruction and prevent dangling callbacks.

### Testing

  `livekit_client`): rapid connect/hang-up cycles and swipe-kill teardown no
  longer crash on the event channel.

Fixes #2127

### DIFF
--- a/common/cpp/src/flutter_common.cc
+++ b/common/cpp/src/flutter_common.cc
@@ -100,9 +100,9 @@ class EventChannelProxyImpl : public EventChannelProxy {
  
      channel_->SetStreamHandler(std::move(handler));
    }
- 
-   virtual ~EventChannelProxyImpl() {}
- 
+
+   virtual ~EventChannelProxyImpl() { channel_->SetStreamHandler(nullptr); }
+
    void Success(const EncodableValue& event, bool cache_event = true) override {
      if (on_listen_called_) {
        PostEvent(event);
--- a/common/cpp/src/flutter_peerconnection.cc
+++ b/common/cpp/src/flutter_peerconnection.cc
@@ -395,10 +395,6 @@ void FlutterPeerConnection::RTCPeerConnectionClose(
     base_->peerconnections_.erase(it2);
   }
 
-  auto it = base_->peerconnection_observers_.find(uuid);
-  if (it != base_->peerconnection_observers_.end())
-    base_->peerconnection_observers_.erase(it);
-
   result->Success();
 }
 
@@ -406,6 +402,10 @@ void FlutterPeerConnection::RTCPeerConnectionDispose(
     RTCPeerConnection* pc,
     const std::string& uuid,
     std::unique_ptr<MethodResultProxy> result) {
+  auto it = base_->peerconnection_observers_.find(uuid);
+  if (it != base_->peerconnection_observers_.end())
+    base_->peerconnection_observers_.erase(it);
+
   result->Success();
 }
 


cloudwebrtc (approved):
lgtm!

PR #2133, opened 2026-07-29
fix(linux): guard texture proxy and harden desktop source enumeration

### LINKED ISSUE #2129: Linux desktop screen capture crashes: texture proxy UAF + fragile source enumeration
**Describe the bug**

Linux desktop screen capture has two crash/fragility issues in the common C++ layer and the GTK texture path.

**(A) Use-after-free in the GTK texture proxy (`linux/flutter/core_implementations.cc`)**

`fl_texture_proxy_copy_pixels` does an unguarded `std::get<flutter::PixelBufferTexture>(*proxy->texture)`. If the texture variant has been destroyed before the proxy `GObject` is torn down (a race during unregister), this dereferences freed memory and crashes. Additionally, `UnregisterTexture` calls `fl_texture_registrar_unregister_texture` *before* nulling `proxy->texture`, leaving a window in which the engine's copy-pixels callback can fire against a destroyed variant.

**(B) Fragile desktop source enumeration (`common/cpp/src/flutter_screen_capture.cc`)**

`BuildDesktopSourcesList` calls `source_list->UpdateSourceList(force_reload)` with no exception guard. On Linux this can throw (e.g. during transient desktop-enumerator failures), aborting the whole enumeration. Separately, `GetDisplayMedia` errors out with `"source not found!"` if the requested `source_id` is not present in the cached list, even when the list is simply stale — a re-enumeration would have found it.

A couple of leftover commented-out `std::cout` debug statements are also present in this file.

**To Reproduce**

1. Linux desktop (GTK).
2. (A) Rapidly start/stop a screen-share capture and tear down the texture; intermittently crash in `fl_texture_proxy_copy_pixels` during the unregister race.
3. (B) Call `getDisplayMedia` with a `sourceId` that is valid but absent from a stale source list (e.g. after the desktop session changed), or trigger a desktop-enumerator transient failure during `UpdateSourceList`; observe a crash or a spurious "source not found" error.

**Expected behavior**

(A) The texture proxy null-checks `proxy->texture` before `std::get`, and `UnregisterTexture` nulls `proxy->texture` *before* calling `fl_texture_registrar_unregister_texture`, closing the race. (B) `UpdateSourceList` is wrapped in `try/catch` on Linux so a transient failure skips that source type instead of aborting, and `GetDisplayMedia` rebuilds the source list and retries before giving up.

**Platform information**

* **OS:** Linux (Desktop / GTK)
* **OS version:** Fedora 43

**Proposed fix**

- `linux/flutter/core_implementations.cc`: add `if (proxy->texture == nullptr) return TRUE;` guard at the top of `fl_texture_proxy_copy_pixels`; null `FL_TEXTURE_PROXY(texture)->texture` before `fl_texture_registrar_unregister_texture` in `UnregisterTexture`.
- `common/cpp/src/flutter_screen_capture.cc`: wrap `UpdateSourceList` in `try/catch` on `__linux__`; in `GetDisplayMedia`, add a Linux fallback that rebuilds the source list and retries when the requested `source_id` is not found; remove the leftover debug `std::cout` statements.

---

Quantumheart:
fix(linux): guard texture proxy and harden desktop source enumeration (#2129)

**Problem**

Linux desktop screen capture has two crash/fragility issues.

**(A) Use-after-free in the GTK texture proxy** (`linux/flutter/core_implementations.cc`)

`fl_texture_proxy_copy_pixels` does an unguarded
`std::get<flutter::PixelBufferTexture>(*proxy->texture)`. If the texture
variant is destroyed before the proxy `GObject` is torn down (a race during
unregister), this dereferences freed memory and crashes. `UnregisterTexture`
also calls `fl_texture_registrar_unregister_texture` *before* nulling
`proxy->texture`, leaving a window in which the engine's copy-pixels callback
can fire against a destroyed variant.

**(B) Fragile desktop source enumeration** (`common/cpp/src/flutter_screen_capture.cc`)

`BuildDesktopSourcesList` calls `UpdateSourceList` with no exception guard; on
Linux this can throw during transient desktop-enumerator failures, aborting
the whole enumeration. `GetDisplayMedia` also errors with `"source not found!"`
when the requested `source_id` is absent from a stale cached list, even though
a re-enumeration would have found it.

A number of leftover `std::cout` debug statements were also present.

**Fix**

- `fl_texture_proxy_copy_pixels`: null-check `proxy->texture` before `std::get`
  (return `TRUE` to avoid crashing the engine copy path).
- `UnregisterTexture`: null `FL_TEXTURE_PROXY(texture)->texture` *before*
  `fl_texture_registrar_unregister_texture`, closing the race.
- `BuildDesktopSourcesList`: on `__linux__`, wrap `UpdateSourceList` in
  `try/catch` so a transient failure skips that source type instead of aborting.
- `GetDisplayMedia`: on `__linux__`, when the requested `source_id` is not in the
  cached list, rebuild the list and retry the match. The first-available
  (`sources_.front()`) fallback is only used when the caller did not specify a
  source (`source_id == "0"`); a specifically-requested source that still isn't
  found falls through to the existing `"source not found!"` error rather than
  silently capturing the wrong surface.
- Remove the leftover debug `std::cout` statements.

### Testing

  `livekit_client`): rapid start/stop of screen-share capture no longer crashes
  during the texture unregister race, and transient desktop-enumerator failures
  no longer abort source enumeration.

Fixes #2129

### DIFF
--- a/common/cpp/src/flutter_screen_capture.cc
+++ b/common/cpp/src/flutter_screen_capture.cc
@@ -1,5 +1,7 @@
 #include "flutter_screen_capture.h"
 
+#include <stdexcept>
+
 namespace flutter_webrtc_plugin {
 
 FlutterScreenCapture::FlutterScreenCapture(FlutterWebRTCBase* base)
@@ -17,7 +19,6 @@ bool FlutterScreenCapture::BuildDesktopSourcesList(const EncodableList& types,
     } else if (type_str == "window") {
       desktop_type = DesktopType::kWindow;
     } else {
-      // std::cout << "Unknown type " << type_str << std::endl;
       return false;
     }
     scoped_refptr<RTCDesktopMediaList> source_list;
@@ -29,7 +30,15 @@ bool FlutterScreenCapture::BuildDesktopSourcesList(const EncodableList& types,
       source_list->RegisterMediaListObserver(this);
       medialist_[desktop_type] = source_list;
     }
-    source_list->UpdateSourceList(force_reload);
+#ifdef __linux__
+    try {
+      source_list->UpdateSourceList(force_reload, false);
+    } catch (...) {
+      continue;
+    }
+#else
+    source_list->UpdateSourceList(force_reload, false);
+#endif
     int count = source_list->GetSourceCount();
     for (int j = 0; j < count; j++) {
       sources_.push_back(source_list->GetSource(j));
@@ -61,7 +70,6 @@ void FlutterScreenCapture::GetDesktopSources(
     sources.push_back(EncodableValue(info));
   }
 
-  //std::cout << " sources: " << sources.size() << std::endl;
   auto map = EncodableMap();
   map[EncodableValue("sources")] = sources;
   result->Success(EncodableValue(map));
@@ -81,9 +89,6 @@ void FlutterScreenCapture::UpdateDesktopSources(
 
 void FlutterScreenCapture::OnMediaSourceAdded(
     scoped_refptr<MediaSource> source) {
-  std::cout << " OnMediaSourceAdded: " << source->id().std_string()
-            << std::endl;
-
   EncodableMap info;
   info[EncodableValue("event")] = "desktopSourceAdded";
   info[EncodableValue("id")] = EncodableValue(source->id().std_string());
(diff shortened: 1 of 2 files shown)


cloudwebrtc (approved):
lg!

Issue #1789, opened 2025-03-17
Webrtc Encryption not working in iOS/Macos


JayS0223:
**When I tried to use the End to End  Encryption in the Flutter_webrtc example, in Android it is working fine, but when it comes to iOS & Macos the app is crashing unexpectedly. I am trying to run sdk example.**

the screen is getting stuck when tried the example for encryption. 

flutter_webrtc: 0.12.12+hotfix.1

https://github.com/user-attachments/assets/2fd1d822-b124-4779-9d02-243a0d71b120

Issue #1325, opened 2023-04-27
Crashing the device when FrameCryptorStateDecryptionFailed


MohsenHaydari:
If I enter the decryption key wrongly, the devices will crash

I use [this tutorial](https://github.com/flutter-webrtc/flutter-webrtc/blob/main/example/lib/src/device_enumeration_sample.dart) to communicate between two devices

I put the _enableEncryption method after creating the RTCPeerConnection and the _enableDecryption method after the _onAddStream is called, if the decryption and encryption keys are correct, everything works correctly, but if the keys are different, the devices will crash. I tested Android 12 emulator and Samsung Android 8  (real device)

```
KeyProvider? _keyManager;
  Future<void> _enableEncryption(List<RTCRtpSender> senders) async {
    _keyManager ??= await _frameCyrptorFactory.createDefaultKeyProvider(_keyProviderOptions);
    for (final element in senders) {
      final String id = '${element.track?.kind}_${element.track?.id}_sender';
      final frameCyrptor = await _frameCyrptorFactory.createFrameCryptorForRtpSender(
          participantId: id, sender: element, algorithm: Algorithm.kAesGcm, keyProvider: _keyManager!);
      frameCyrptor.onFrameCryptorStateChanged = (participantId, state) {
        log("state $state");
      };

      await frameCyrptor.setKeyIndex(0);
      await frameCyrptor.setEnabled(true);
      await _keyManager!.setKey(participantId: id, index: 0, key: aesKey);
      for (final i in allCodec) {
        await frameCyrptor.updateCodec(i);
      }
    }
  }

  Future<void> _enableDecryption(List<RTCRtpReceiver> receivers) async {
    _keyManager ??= await _frameCyrptorFactory.createDefaultKeyProvider(_keyProviderOptions);
    for (var element in receivers) {
      final String id = '${element.track?.kind}_${element.track?.id}_sender';
      final frameCyrptor = await _frameCyrptorFactory.createFrameCryptorForRtpReceiver(
          participantId: id, receiver: element, algorithm: Algorithm.kAesGcm, keyProvider: _keyManager!);
      frameCyrptor.onFrameCryptorStateChanged = (participantId, state) {
        log("DecryptState $state");
      };

      await frameCyrptor.setKeyIndex(0);
      await frameCyrptor.setEnabled(true);
      await _keyManager!.setKey(participantId: id, index: 0, key: recAes); // diffrenet key
      for (final i in allCodec) {
[...]

Issue #1249, opened 2023-01-25
WebRtc encryption in Flutter


wilmakelcey:
Hello friends
Please answer, it is very important to me

1- Is it possible to have arbitrary encryption in Webrtc in the web or mobile version?
If possible, how is it?

2- If you cannot have a desired encryption, how can you change the encryption key?
For example, the page: https://webrtc.github.io/samples/src/content/insertable-streams/endtoend-encryption/
This is possible
How can I change the encryption key as desired?

These 2 questions are very important to me and I have been having problems in my program for a long time, please guide me

Thank you very much

cloudwebrtc:
This is the currently implemented frame encryption feature
https://github.com/flutter-webrtc/flutter-webrtc/pull/1201

Use aes-gcm to implement custom encryption. It is currently expected to merge and release a new version within 1 week. You can implement end-to-end encryption by setting a custom key for each track.

PR #1201, opened 2022-12-09
feat: framecryptor.

cloudwebrtc:
- [x] iOS
- [x] macOS
- [x] Android
- [x] Windows
- [x] Web
- [x] Linux

PR #2135, opened 2026-08-01
fix(darwin): deliver FrameCryptor state events reliably and fail safe on decrypt failure

brennodewinter:
## Summary

FrameCryptor (frame-level E2EE) works on Android but users report it crashing / hanging on iOS & macOS (#1789, #1325). This PR makes the **Darwin** FrameCryptor state-delivery path match the (non-crashing) **Android** implementation and fail safe on a decryption failure.

> **Honesty note for reviewers:** this is a **code-level parity/hardening fix derived by comparing the Darwin and Android bridges**, not one confirmed against a live symbolicated crash trace. I was unable to stand up a two-peer E2EE session on real Apple hardware to force a decrypt failure and capture a trace. The reporters in #1789/#1325 did not attach traces either. The reasoning is below so it can be scrutinised; please build/verify against the example before merging.

## Root cause

The task's usual suspect — emitting the event **off the platform thread** — is **already fixed on `main`**: `postEvent()` wraps delivery in `dispatch_async(dispatch_get_main_queue(), …)` and nil-guards the sink (`FlutterWebRTCPlugin.m`, landed in #1458). So that is not the remaining bug.

Comparing `common/darwin/Classes/FlutterRTCFrameCryptor.m` against `android/.../FlutterRTCFrameCryptor.java`, three divergences remain on Darwin:

1. **State events are dropped instead of queued.** Android buffers events in an `eventQueue` when no listener is attached yet and flushes them in `onListen` (`AnyThreadSink` + `eventQueue`). On Darwin the delegate simply does nothing when `eventSink == nil`. The Dart side attaches its `EventChannel.receiveBroadcastStream().listen(...)` **after** the native factory call returns (`lib/src/native/frame_cryptor_impl.dart`, `FrameCryptorImpl` constructor), so any state that fires in that window — notably an immediate `decryptionFailed` / `missingKey` on a wrong/rotated key — is lost. This matches the "E2EE doesn't work / screen stuck" symptom.

2. **`nil` `participantId` crashes the process.** In `-frameCryptor:didStateChangeWithParticipantId:withState:` the value was inserted directly into an `NSDictionary` literal. A nil value raises `NSInvalidArgumentException` ("attempt to insert nil object"), on the delegate thread.

3. **No teardown on dispose.** `frameCryptorDispose` only removed the entry and set `enabled = NO`. It never cleared the delegate or the stream handler, so the native cryptor keeps delivering state events to a disposed object, and `eventChannel` (retained as an associated object of the cryptor) and the cryptor (retained as the channel's stream handler) form a retain cycle. Android calls `frameCryptor.dispose()`.

## Fix

Mirror Android in `common/darwin/Classes/FlutterRTCFrameCryptor.{h,m}`:

- Add an `eventQueue` on the `RTCFrameCryptor (Flutter)` category; buffer state events when no sink is attached and **flush them in `-onListenWithArguments:`**.
- **Guard `participantId`** (`participantId ?: @""`) before building the event dictionary.
- On dispose, **tear down**: `delegate = nil`, `setStreamHandler:nil`, drop the channel/sink, clear the queue — stopping stray events and breaking the retain cycle.
- Guard the cross-thread `eventQueue` with `@synchronized` (it is mutated on the WebRTC worker thread and read on the main thread).

The **failure state still reaches Dart** — only the crash and the lost-event window are removed. No AES-GCM/crypto math is touched and there is **no change to the prebuilt libwebrtc binary**.

## Diff

`common/darwin/Classes/FlutterRTCFrameCryptor.h` (+4), `common/darwin/Classes/FlutterRTCFrameCryptor.m` (+47/−5). No formatting churn.

## Testing / verification status

- ✅ Static comparison against the Android implementation for each transition (`new`, `ok`, `encryptionFailed`, `decryptionFailed`, `missingKey`, `keyRatcheted`, `internalError`).
- ✅ All `RTCFrameCryptorState` cases handled in `stringFromState:`.
- ⚠️ **Not** verified against a live macOS/iOS E2EE call or a symbolicated trace (see honesty note). Suggested manual test for a reviewer with Apple hardware:
  1. Run `example/` E2EE sample on macOS + an iOS device, two peers.
  2. Matching keys → confirm audio/video decrypts (parity with Android).
  3. Set a mismatched/rotated key on one side → confirm the app **no longer crashes** and a `FrameCryptorState.decryptionFailed` (or `missingKey`) reaches the Dart `onFrameCryptorStateChanged` callback.
  4. Re-run the same path on Android → **no regression**.

## Related

Relates to #1789, #1325. Context: #300, #1249. Darwin FrameCryptor was added in #1201.

### DIFF
--- a/common/darwin/Classes/FlutterRTCFrameCryptor.h
+++ b/common/darwin/Classes/FlutterRTCFrameCryptor.h
@@ -11,6 +11,10 @@
 @interface RTCFrameCryptor (Flutter) <FlutterStreamHandler>
 @property(nonatomic, strong, nullable) FlutterEventSink eventSink;
 @property(nonatomic, strong, nullable) FlutterEventChannel* eventChannel;
+// Buffers state-change events fired by the native cryptor before the Dart side
+// has attached its event-channel listener, so they are not lost. Mirrors the
+// Android implementation's eventQueue. Flushed in -onListenWithArguments:.
+@property(nonatomic, strong, nullable) NSMutableArray* eventQueue;
 @end
 
 
--- a/common/darwin/Classes/FlutterRTCFrameCryptor.m
+++ b/common/darwin/Classes/FlutterRTCFrameCryptor.m
@@ -22,6 +22,15 @@
                            OBJC_ASSOCIATION_RETAIN_NONATOMIC);
 }
 
+- (NSMutableArray*)eventQueue {
+  return objc_getAssociatedObject(self, _cmd);
+}
+
+- (void)setEventQueue:(NSMutableArray*)eventQueue {
+  objc_setAssociatedObject(self, @selector(eventQueue), eventQueue,
+                           OBJC_ASSOCIATION_RETAIN_NONATOMIC);
+}
+
 #pragma mark - FlutterStreamHandler methods
 
 - (FlutterError* _Nullable)onCancelWithArguments:(id _Nullable)arguments {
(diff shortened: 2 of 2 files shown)


cloudwebrtc (approved):
lgtm!

PR #2145, opened 2026-08-11
perf(windows): run AddTransceiver/AddTrack off the platform thread

ErtasBurak:
### Problem

`pc->AddTransceiver()` and `AddTrack()` are proxy calls — they **block the caller** until the signaling thread has finished. On Windows the Flutter platform thread is the UI message loop, so publishing a track blocks the whole interface: the app stops painting and the cursor stops responding.

In my app, joining a voice call was a visible freeze of roughly **600 ms**, most of it the *first* audio transceiver — that is the call that brings the transport up.

### Change

Move the blocking part to a detached worker, following the pattern this file already uses for `CreateOffer`/`CreateAnswer` rather than introducing a new one.

The constraints map is parsed on the calling thread first, so the worker captures only refcounted objects (`pc`, the track, the parsed init, the result proxy) — it needs neither the `EncodableMap` nor the `FlutterPeerConnection` instance, and so cannot outlive anything it touches.

### Safety

- The proxy is thread-safe: it marshals to the signaling thread regardless of which thread calls it.
- The result is completed from the worker, so **ordering as seen by Dart is unchanged** — the caller still awaits exactly one completion.
- Guarded by `#ifdef _WIN32`, since that is where the platform thread is the message loop and where the freeze is visible. Other platforms keep the current path.

Happy to widen it beyond Windows, or to fold it behind the same helper as the offer/answer path, if you would prefer either.

### DIFF
--- a/common/cpp/src/flutter_peerconnection.cc
+++ b/common/cpp/src/flutter_peerconnection.cc
@@ -6,6 +6,12 @@
 #include "rtc_dtmf_sender.h"
 #include "rtc_rtp_parameters.h"
 
+#ifdef _WIN32
+// Worker thread for off-platform-thread track publish
+// (see AddTransceiver / AddTrack).
+#include <thread>
+#endif
+
 namespace flutter_webrtc_plugin {
 
 std::string RTCMediaTypeToString(RTCMediaType type) {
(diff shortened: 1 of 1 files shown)


cloudwebrtc (approved):
lgtm!

PR #2152, opened 2026-08-20
fix(windows): don't unregister event channel handlers after engine teardown

### LINKED ISSUE #2151: [Windows] Crash on every exit: ~EventChannelProxyImpl unregisters its stream handler after the engine has detached the messenger
I'm seeing a crash on exit when using flutter_webrtc 1.6.0 on Windows 11 with Flutter 3.44.4 stable. This happens even if I never actually use the widget; just depending on the plugin triggers it. It happens every time, but may not be noticeable to some users because it doesn't show a crash dialogue. I only noticed because the process lingers for 15 to 20 seconds while Windows makes a crash dump, and I was waiting for its exit from another process.

This is a regression introduced in 1.6.0 in #2131 (which was to fix #2127).

I used an LLM (Claude Fable 5) to track the problem down to flutter_common.cc:148:

```cpp
virtual ~EventChannelProxyImpl() { channel_->SetStreamHandler(nullptr); }
```

The text below is from a Claude Fable 5 and describes the problem and suggested fix, which works for me (apologies for the LLM generated text).

---

`EventChannelProxyImpl::~EventChannelProxyImpl()` (common/cpp/src/flutter_common.cc:148) calls `channel_->SetStreamHandler(nullptr)`. During engine shutdown this destructor runs from `PluginRegistrar::ClearPlugins()` *after* `FlutterWindowsEngine` has already detached itself from the `FlutterDesktopMessenger`, so the call reaches `FlutterDesktopMessengerSetCallback`, which dereferences the detached (null) engine — an access violation (0xc000041d) on the platform thread. The engine documents this as a caller-side contract: the check `FlutterDesktopMessengerIsAvailable(messenger)` is only an `FML_DCHECK` ("Messenger must reference a running engine to set a callback"), compiled out in release builds. Registering the plugin is sufficient to hit this — no plugin API needs to be called — because the proxy is created eagerly at registration (`FlutterWebRTCBase` constructor) and destroyed at engine teardown.

To fix, unregister the stream handler only when the messenger still references a running engine, using the engine's own liveness API (`FlutterDesktopMessengerAddRef` / `Lock` / `IsAvailable` / `Unlock` / `Release` from `flutter_messenger.h`). Mechanical steps against v1.6.0:

1. **Capture the raw messenger at registration** — in `windows/flutter_webrtc_plugin.cc`, `FlutterWebRTCPluginRegisterWithRegistrar` already receives the raw `FlutterDesktopPluginRegistrarRef`. Obtain `FlutterDesktopMessengerRef raw = FlutterDesktopPluginRegistrarGetMessenger(registrar);` and pass it down (ownership: each proxy AddRefs it on construction, see step 3) (e.g. extend `FlutterWebRTCPluginImpl::RegisterWithRegistrar` and the private constructor with a `FlutterDesktopMessengerRef` parameter, stored as a member and exposed via a `raw_messenger()` accessor on `FlutterWebRTCPlugin`, alongside the existing `messenger()` / `textures()` / `task_runner()` accessors).

2. **Thread it to the proxy** — `FlutterWebRTC` / `FlutterWebRTCBase` constructor gains the ref as an extra argument, stored on `FlutterWebRTCBase`, and `EventChannelProxy::Create` gains a 4th parameter (defaulting to `nullptr`) passed at **all four call sites**: the base event channel (flutter_webrtc_base.cc:23), per peer connection (flutter_peerconnection.cc:1182), per data channel (flutter_data_channel.cc:12), and per video renderer (flutter_video_renderer.cc:18).

3. **Guarded destructor** — in `common/cpp/src/flutter_common.cc`, `EventChannelProxyImpl` stores the ref and replaces the destructor with the block below. Refcount discipline: since several proxies exist (base channel plus one per peer connection / data channel / video renderer), each proxy must take its OWN reference — call `FlutterDesktopMessengerAddRef(raw)` in the proxy constructor — so the per-destructor `Release` below is balanced. (A single shared AddRef with per-proxy Releases would over-release.)

   ```cpp
   virtual ~EventChannelProxyImpl() {
     if (raw_messenger_) {
       // Unregister only while the engine is alive. During engine teardown
       // the messenger is already detached and SetStreamHandler(nullptr)
       // would reach FlutterDesktopMessengerSetCallback, which dereferences
       // the detached engine (release builds compile out its FML_DCHECK):
       // access violation on the platform thread. The registration dies
       // with the engine anyway.
       FlutterDesktopMessengerLock(raw_messenger_);
       if (FlutterDesktopMessengerIsAvailable(raw_messenger_)) {
         channel_->SetStreamHandler(nullptr);
       }
       FlutterDesktopMessengerUnlock(raw_messenger_);
       FlutterDesktopMessengerRelease(raw_messenger_);
     } else {
       channel_->SetStreamHandler(nullptr);  // non-Windows: unchanged
     }
   }
   ```

4. **Linux/shared-code note** — `common/cpp` is shared with the Linux build, which has no `FlutterDesktopMessenger` API; there the new parameter stays `nullptr` and behaviour is unchanged (the `raw_messenger_` member can be `void*` or hidden behind a small platform typedef/#ifdef to avoid including `flutter_messenger.h` in shared headers).

This preserves legitimate unregistration whenever a plugin instance is destroyed while its engine is alive, and no-ops during engine teardown, where the registration is dying with the engine regardless.

[flutter_webrtc_guarded_fix.patch](https://github.com/user-attachments/files/31241475/flutter_webrtc_guarded_fix.patch)

cloudwebrtc:
`~EventChannelProxyImpl` unconditionally called
`channel_->SetStreamHandler(nullptr)`. On Windows that destructor also runs from `PluginRegistrar::ClearPlugins()` during `FlutterWindowsEngine` teardown, which happens *after* the engine has detached itself from the `FlutterDesktopMessenger`. The call then reaches
`FlutterDesktopMessengerSetCallback`, which dereferences the detached engine — its `FML_DCHECK` that the messenger must reference a running engine is compiled out in release builds — and the process dies with an access violation (0xc000041d) on the platform thread on every exit.

Merely depending on the plugin is enough to hit this: the base event channel proxy is created eagerly when `FlutterWebRTCBase` is constructed at registration, so it is always alive at engine teardown.

Capture the host messenger at plugin registration and have each event channel proxy hold its own reference to it (`AddRef`/`Release`, so the handle outlives the engine and the refcount stays balanced across the base channel and the per-peer-connection / data-channel / video-renderer channels). The destructor now unregisters only while `FlutterDesktopMessengerIsAvailable` reports a live engine under the messenger lock; during teardown the registration dies with the engine anyway. Non-Windows builds of the shared `common/cpp` code keep the previous unguarded behaviour.

Regression from #2131. Fixes #2151.

### DIFF
--- a/common/cpp/include/flutter_common.h
+++ b/common/cpp/include/flutter_common.h
@@ -186,4 +186,12 @@ class EventChannelProxy {
                        bool cache_event = true) = 0;
 };
 
+// Records the host (embedder) messenger handle captured at plugin
+// registration. It is used to check that the engine is still running before
+// an event channel unregisters its stream handler during teardown, see
+// ~EventChannelProxyImpl. The handle is opaque here so this shared header
+// stays platform-neutral; platforms without the FlutterDesktopMessenger C API
+// never call this and keep the unguarded behaviour.
+void SetEventChannelHostMessenger(void* messenger_ref);
+
 #endif  // FLUTTER_WEBRTC_COMMON_HXX
--- a/common/cpp/src/flutter_common.cc
+++ b/common/cpp/src/flutter_common.cc
@@ -3,6 +3,19 @@
 
 #include <memory>
 
+#if defined(_WIN32)
+#include <flutter_messenger.h>
+#endif
+
+// See SetEventChannelHostMessenger() in flutter_common.h. Only ever written
+// from plugin registration, which happens on the platform thread before any
+// event channel exists.
+static void* g_host_messenger = nullptr;
+
+void SetEventChannelHostMessenger(void* messenger_ref) {
+  g_host_messenger = messenger_ref;
+}
+
 class MethodCallProxyImpl : public MethodCallProxy {
  public:
   explicit MethodCallProxyImpl(const MethodCall& method_call)
@@ -122,6 +135,14 @@ class EventChannelProxyImpl : public EventChannelProxy {
              channelName,
              &flutter::StandardMethodCodec::GetInstance())),
              task_runner_(task_runner) {
+#if defined(_WIN32)
+     // Each proxy keeps its own reference so that the handle outlives the
+     // engine it belongs to and the Release() below stays balanced.
+     if (g_host_messenger) {
+       host_messenger_ = FlutterDesktopMessengerAddRef(
+           static_cast<FlutterDesktopMessengerRef>(g_host_messenger));
+     }
+#endif
      auto handler = std::make_unique<
          flutter::StreamHandlerFunctions<EncodableValue>>(
          [&](const EncodableValue* arguments,
@@ -145,7 +166,29 @@ class EventChannelProxyImpl : public EventChannelProxy {
      channel_->SetStreamHandler(std::move(handler));
    }
 
-   virtual ~EventChannelProxyImpl() { channel_->SetStreamHandler(nullptr); }
+   virtual ~EventChannelProxyImpl() {
+     // Unregister the stream handler only while the messenger still
+     // references a running engine. This destructor also runs from
+     // PluginRegistrar::ClearPlugins() during FlutterWindowsEngine teardown,
+     // after the engine has detached itself from the FlutterDesktopMessenger;
+     // an unconditional SetStreamHandler(nullptr) then reaches
+     // FlutterDesktopMessengerSetCallback, which dereferences the detached
+     // engine (its FML_DCHECK is compiled out in release builds): access
+     // violation on the platform thread. The registration dies with the
+     // engine anyway.
+#if defined(_WIN32)
+     if (host_messenger_) {
+       FlutterDesktopMessengerLock(host_messenger_);
+       if (FlutterDesktopMessengerIsAvailable(host_messenger_)) {
+         channel_->SetStreamHandler(nullptr);
+       }
+       FlutterDesktopMessengerUnlock(host_messenger_);
+       FlutterDesktopMessengerRelease(host_messenger_);
+       return;
+     }
+#endif
+     channel_->SetStreamHandler(nullptr);
+   }
 
    void Success(const EncodableValue& event, bool cache_event = true) override {
      if (on_listen_called_) {
@@ -177,6 +220,9 @@ class EventChannelProxyImpl : public EventChannelProxy {
    std::list<EncodableValue> event_queue_;
    bool on_listen_called_ = false;
    TaskRunner* task_runner_;
+#if defined(_WIN32)
+   FlutterDesktopMessengerRef host_messenger_ = nullptr;
+#endif
  };
 
 std::unique_ptr<EventChannelProxy> EventChannelProxy::Create(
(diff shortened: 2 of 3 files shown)


cloudwebrtc:
Hey @arthur-tacca, could you take a look at this PR? I'm not sure here: https://github.com/flutter-webrtc/flutter-webrtc/pull/2152/changes#diff-8d7b1bb9613915878d3079ff186a911fbd100c9568a5c7099df15bbbd5048599R83-R84, `FlutterDesktopMessengerAddRef` ref count needs to be incremented once, which should cause a leak if `FlutterDesktopMessengerRelease` is not called during the destruction of `FlutterWebRTCPluginImpl`.

arthur-tacca:
@cloudwebrtc This is way deep inside a stack that I know nothing about, but I can ask an LLM about it. This is more useful than it sounds, as it's already primed with context from finding the bug. Below is its output.

---

Their implementation is the shape I'd endorse merging — it fixes the exit crash, preserves the #2131 mid-session unregistration, and keeps `common/cpp` clean for Linux.

**The answer to their question: there is no leak — the accounting in the PR as written is exactly balanced.** The key is that the registration-time global stores a borrowed pointer with no reference of its own (`SetEventChannelHostMessenger(FlutterDesktopPluginRegistrarGetMessenger(registrar))` — no `AddRef` there). The only `AddRef`s are per-proxy, in `EventChannelProxyImpl`'s constructor, and each one is paired with the Release in that same proxy's destructor. Every proxy is destroyed eventually — the base channel proxy inside the `FlutterWebRTCPluginImpl → FlutterWebRTC → FlutterWebRTCBase` destruction chain at engine teardown, and the per-peer-connection/data-channel/renderer proxies whenever their owners die — so every `AddRef` meets its `Release`. Nothing needs releasing in `~FlutterWebRTCPluginImpl` itself, because the plugin object never took a reference. (And releasing a handle of an already-detached messenger is safe by design — the messenger is a refcounted object built to outlive its engine; our instrumented test runs exercised exactly this `Lock`/`IsAvailable`/`Unlock`/`Release` sequence on a detached messenger, three for three.)

Two genuinely minor caveats worth them noting in a comment rather than changing code:

1. After an engine dies and its last proxy releases, `g_host_messenger` dangles — but it's harmless because the global is only ever read in proxy constructors, which only run from plugin code under a live engine, and any new engine's registration overwrites the global before its first proxy exists.
2. The narrow multi-engine window: if two engines were alive simultaneously, a proxy created for engine A after engine B registered would capture B's messenger and guard against the wrong engine's liveness. Sequential engine restarts are fine; concurrent multi-engine would need the ref threaded per-registrar instead of the global. For this plugin's realistic usage that's a documentation footnote, not a blocker.

arthur-tacca:
I hadn't seen that there were two commits to this PR. Claude's answer above, that AddRef and Release are exactly balanced, was based on the overall PR with both commits i.e. it agrees with Copilot's concern and its suggested fix.

I also checked it empirically on my machine - both as a dependency without instantiating (which was my original problem case) and actually including it as a widget (loopback of my own screen, with mid-session hangup), and it no longer crashes on exit.

PR #2179, opened 2026-09-14
fix(darwin): stop leaking every platform view that is disposed

gfx:
An app that renders video through the platform view path never frees a
renderer. Every view it has shown is still alive, holding its native view and
the last frame drawn into it: one macOS client had 25 of them for 14 live video
tracks, and 146 MB across 186 IOSurface regions.

`setStreamHandler:` makes the binary messenger retain the controller and
nothing ever clears it, so taking it out of the factory's dictionary frees
nothing. `videoPlatformViewRendererDispose` can clear it, because Dart cancels
the event channel subscription before invoking that method.

#2167 fixed this for the texture path and missed the platform view beside it.


cloudwebrtc (approved):
lgtm!

Issue #2156, opened 2026-08-27
RTCDataChannel close leaves stale open state and allows sends past native teardown


zhuhaichao518:
**Describe the bug**

`RTCDataChannelNative.close()` does not establish a Dart-side terminal state or send fence. It cancels the event subscription and removes the native DataChannel mapping through `dataChannelClose`, but leaves the cached Dart `state` unchanged. A channel that was open can therefore still report `RTCDataChannelOpen` after close has started or completed.

A concurrent or late `send()` then crosses the native teardown boundary and fails from the platform implementation rather than being rejected locally:

```text
PlatformException(
  dataChannelSendFailed,
  dataChannelSend() data_channel is null
)
```

On the desktop C++ implementation, `DataChannelClose` removes the UUID from `data_channel_observers_`, while `send()` performs a later UUID lookup. There is no Dart `_closeRequested` fence and no retained native reference spanning a competing lookup/removal.

**To Reproduce**

1. Establish and open a native DataChannel.
2. Start an asynchronous producer that can have a send pending or scheduled.
3. Call `close()` while that producer is active.
4. Observe that the Dart object may still report `RTCDataChannelOpen` and a late/concurrent send reaches native code after its UUID mapping has been removed.
5. On Windows, observe `dataChannelSendFailed: data_channel is null`.

A simpler state demonstration is to inspect `dc.state` after a locally initiated `await dc.close()`; because event delivery is cancelled and `close()` never updates `_state`, it can retain the previous open state.

**Expected behavior**

Starting close should synchronously establish a terminal send fence and transition the Dart object to closing/closed. New sends should be rejected locally with a defined result, already-started sends should settle safely, repeated close should be idempotent, and native calls should retain/synchronize channel ownership across lookup and teardown.

Related downstream investigation: https://github.com/CloudPlayPlus/cloudplayplus/issues/617

**Platform information**

* **Flutter version**: 3.41.6
* **Plugin version**: 1.6.0 and current `main`
* **OS**: Windows Desktop
* **OS version**: Observed in Release builds; not isolated to a specific Windows version

PR #2191, opened 2026-09-20
fix: establish a Dart-side terminal state and send fence on data channel close

cipherprofessor:
Contributes to #2156 (does not close it — see Scope below).

## Problem

`RTCDataChannelNative.close()` never updated the cached Dart `state` or notified listeners of it. It cancels the event subscription and removes the native DataChannel mapping through `dataChannelClose`, but `dc.state`, `stateChangeStream`, and `onDataChannelState` could all keep reporting the previous open state after `close()` completed.

`send()` had no local guard either, so a call issued after `close()` reached the platform channel and failed there instead of being rejected locally:

```
PlatformException(dataChannelSendFailed, dataChannelSend() data_channel is null)
```

## Fix

- `close()` now sets `_state` to `RTCDataChannelClosed` and notifies both `onDataChannelState` and `stateChangeStream` of it — mirroring exactly the notification the native-event path (`eventListener`'s `dataChannelStateChanged` handling) already does — before tearing down the stream controllers.
- `send()` now throws a local `Exception` when `_isClosed`, matching this codebase's existing idiom for state-guard errors (e.g. `rtc_video_renderer_impl.dart`'s `"Can't be muted: ..."` checks), so a post-close send never reaches the platform channel at all.

## Scope — please read before reviewing

This addresses the **Dart-layer half** of #2156 only. The issue also describes a genuine race in the **native Windows desktop C++ implementation**: `DataChannelClose` removes a channel's UUID from `data_channel_observers_` while `send()` performs a later UUID lookup, with no retained native reference spanning the two. That's a structurally separate fix in different, native code this PR does not touch — I don't have the native-side context to attempt it safely, and didn't want to hold up the Dart-side fix waiting on it. Filing under "contributes to," not "fixes," so the issue stays open for that part.

## Testing

Added to `test/unit/rtc_data_channel_close_send_test.dart`, following the mocking conventions already established in `test/unit/rtc_peerconnection_dispose_data_channels_test.dart`:
- `close()` transitions `state` to `RTCDataChannelClosed`.
- `send()` after `close()` is rejected locally — asserted both by the thrown exception and by confirming `dataChannelSend` never appears in the platform method-call log.
- `close()` notifies both `stateChangeStream` listeners and `onDataChannelState` of the closed state.

Confirmed the first two tests fail (red) against the pre-fix code before implementing, and pass (green) after.

`flutter test` (full suite), `dart format`, and `flutter analyze` all clean.


cloudwebrtc (approved):
lg

